PR #5, opened 2026-08-05
fix: scope attribution footer validation to the terminal block

ss251:
Port of elizaOS/eliza#17613 (closes the defect elizaOS/eliza#17610 tracks) — the source PR's subject files were deleted from the monorepo when this repository split out, so the fix lands here, where the deployed validator actually lives.

## The defect

`markerFooterError` counts attribution label rows across the WHOLE body. The shipped elizaOS/eliza PR template carries those same rows as its provenance checklist, and the contribution skill instructs appending the machine footer last — so **any contributor following both official documents produces an invalid marker** ("marker requires exactly one complete visible attribution footer"). Measured on the live ledger 2026-08-03: 64 of 67 eligible sources (95.5%) invalid, all with template-shaped bodies.

## The fix

`terminalAttributionBlock` walks label lines back from the terminal lane signature; footer-row counting is scoped to that block. Deliberately unweakened: a bare marker with no lane signature above it, stacked complete footers, and duplicated rows adjacent to the marker all still fail.

## Proof

- New test composes the **shipped** `.github/pull_request_template.md` with the skill's footer — red against the unscoped validator, green here.
- No-weakening case: duplicated footer rows adjacent to the marker still rejected.
- Full suite: **56/56** (`bun test src/lib/leaderboard.test.ts`); biome clean.
- Cross-validated on 2026-08-03 by running both validator builds against real PR bodies via `bun` imports (methodology in elizaOS/eliza#17613's thread).

One test path was adapted from the monorepo layout (`resolve(process.cwd(), ".github/pull_request_template.md")`) — this repo ships its own copy of the template.

---

AI provider/model: Anthropic / claude-fable-5
Contribution skill revision: elizaOS/eliza@e37924817dd457f3c5cc1547531c7054336a3740:packages/skills/skills/contribute-to-eliza
— [claude-code-ss251]


lalalune (approved):
Reviewed the final refreshed head against current develop. The validator now scopes visible attribution rows to the terminal footer block without weakening duplicate-row or multiple-lane rejection. The refresh preserves signed elizaos-contribution v2 receipts, canonical v1 markers, and historical eliza-computer v1 parsing. Local verification on the exact head: 61 focused leaderboard tests, typecheck, and Biome all pass.

PR #28, opened 2026-08-12
feat: profile things that could be worked on

odilitime:
## Summary
- Bump the leaderboard snapshot to schema v5 with deterministic `opportunities` for still-open work (missing/partial evidence, near material-test threshold, expandable reviews).
- Derive monthly cap fill on project contributors and show it as a compact status line under active cycles—not per-drop demerit rows.
- Add a public profile section **Things that could be worked on.** with verb-first hints, `up to +N`, GitHub links, newest-first, max 5.

## Test plan
- [x] `bunx vitest run tests/app.test.tsx src/lib/leaderboard.test.ts src/lib/project-view.test.ts scripts/evidence-contract.test.mjs`
- [x] `bun run typecheck`
- [x] `bun run lint:check` / `bun run format:check`
- [ ] Spot-check `/contributors/<login>` with a fixture or local snapshot: opportunity section appears when open opportunities exist, hidden when empty, cap line shows fill
- [ ] Confirm regenerated `leaderboard.json` is schema `"5"` after `bun run leaderboard:generate`

Made with [Cursor](https://cursor.com)


lalalune (changes requested):
Two merge blockers reproduced against head 155990df56b8: (1) the required test suite fails in scripts/generate-leaderboard.test.ts (current-head review selection) because the fixture publishes a review URL whose fragment does not satisfy the new review-URL validator; the snapshot aborts at snapshot.opportunities[0].source.url. (2) current develop now conflicts in src/App.tsx. Please rebase onto current develop, resolve the conflict without dropping the intervening ASI/ranking work, fix the review URL fixture/contract mismatch, and rerun the complete package and browser checks with a live generated ledger.

lalalune (approved):
Both blockers from the previous review are resolved at head e787537. Noting for the record that I pushed the resolution and fixes to this branch myself, so this approval covers work I authored on top of @odilitime's commits.

**(1) generate-leaderboard.test.ts review-URL failure** — this was a stale-base artifact. The branch's base built review URLs as `#pullrequestreview-REVIEW_STALE`, whose fragment has no digits; develop had already changed that fixture to a numeric `databaseId`. After merging develop the file passes (49/49), so no contract change was needed — real GitHub review URLs are numeric.

**(2) src/App.tsx conflict** — resolved without dropping the intervening work. Both hunks were in `ProfilePage` and the second was semantic, not textual: develop's `85e24f6`/`57748e4` replaced the per-view score summation with `createGlobalLeaders`. I kept **develop's** side for both `events` (from `snapshot.ledger` via `findProjectByRepositoryId`) and `score` (`globalLeader?.score ?? 0`), and carried over only this branch's `opportunities` lookup — folded into the not-found guard and placed after `globalLeader` and `matches` in the actor fallback. Resolving it the other way would have silently reverted the ranking-truthfulness fix on a publicly displayed contributor score.

**Checks** — the complete package and browser checks ran green with a live generated ledger (6m2s). Locally: lint, `tsc -b`, 306 vitest tests, 32 skill-tests, and `bun audit --production` all pass.

### Two additional fixes I made

**`expand-review` published false guidance.** The opportunity loop stops at each author's *first thin* review, while scoring awards on their *first qualifying* one. So a reviewer who left "LGTM" and then a substantive review with inline comments was told to "Add at least 20 characters of review rationale" to earn +3 they were already going to be awarded at merge. Reviewers with a qualifying review on the same PR are now excluded, with a regression test that fails without the change.

**Inverted hedging on point labels.** Evidence was the one category rendered as unconditional ("+6 possible") while everything else read "+N if it qualifies" — despite evidence being the least certain, since each artifact must still pass remote structure and head-binding verification. It now reads "+N if it verifies".

Non-blocking findings from the review — cap-aware `potentialPoints`, public snapshot payload growth, and the validator aborting the whole snapshot on one malformed opportunity row — are filed as #40.

Issue #36, opened 2026-08-13
Skill v2: make Slop contribution flow fast, coherent, and launchable


NubsCarson:
## Outcome

Turn the Slop contributor/reviewer skill family into one coherent, testable work protocol: understand a funded project, select one live bounded outcome, coordinate safely, produce real proof, publish a reviewable contribution, and attach trustworthy provenance without treating token volume as quality.

## Confirmed current gaps

- Installed-style symlink execution of run-receipt.mjs exits 0 without starting; #30 / PR #31 owns that repair, but its current scope omits the newer ASI package.
- live-report.mjs performs many serial GitHub comment/review calls and provides no progress while a large repository is scanned.
- Current provenance coverage is partial: 277 valid sources, 117 missing, and 173 invalid in the fresh 2026-08-13 ledger; relevant token totals are zero.
- Model policy is duplicated and hard-coded per project skill.
- Public source authority and user-facing protocol names still mix Slop with elizaOS/army and gitarmy identifiers; existing receipts, score snapshots, labels, wallet markers, and authorization records require an explicit compatibility migration.
- One scalar score mostly represents capped recent GitHub throughput. Product direction is still required for whether main rank means independently reviewed impact, deterministic throughput, or separate impact/shipping/review/research leagues.

## Acceptance criteria

- Record the chosen ranking/reward meaning and separate quality, provenance, and activity concepts.
- Keep source-verified atomic installation, hostile-input isolation, independent review, immutable reward records, and non-custodial settlement as hard boundaries.
- Make discovery bounded and observably progressing; avoid repository-size N+1 request behavior.
- Make model/client policy declarative from the project contract and validate all project packages from one inventory.
- Design a versioned Slop identity migration that accepts historical Army/gitarmy artifacts without issuing new legacy authority.
- Add installed-layout tests and forward-test a real start → live contribution → finish → GitHub marker → ingestion path with an approved live model.
- Test Eliza, ASI, and Delta Star contributor and reviewer packages, their generated archives, UI install/read-only flows, and failure states.
- Attach reviewed command output, archive/checksum, receipt, GitHub source, live trajectory, browser evidence, and production-applicability notes.

## Coordination

PR #31 remains owned by its author for the receipt-runner repair. Do not duplicate that patch; require it to include ASI and generalized parity coverage. Ranking PR #35 remains independent. No production deploy, DNS, Forgejo, wallet, or settlement lever is included here.

NubsCarson:
[...]

PR #61, opened 2026-08-15
chore: bind private Slop runtime resources

NubsCarson:
## Summary
- bind the existing `slop-private` D1 database and private `slop-private-traces` R2 bucket to Pages
- bind Pages to the private `slop-identity` service and pin the identity Worker to the same D1 database
- remove the partial preview override so Cloudflare does not drop non-inherited bindings
- deploy the identity Worker from the same protected exact-SHA release as Pages
- require the exact four Worker secrets and prove live D1 flow creation, TLS, PKCE S256, callback, and GitHub client routing before Pages publishes
- fail closed with a stable public error if a malformed secret or non-API exception reaches the Worker boundary
- document exact 32-byte base64url key generation and lock the resource/domain/deployment contract in tests

The Cloudflare D1 database, private R2 bucket, migrations, Worker secrets, GitHub App, and restricted GitHub prod-ops runner group now exist. This PR does not enable payouts, expose R2 publicly, grant an operator identity, or install the GitHub App into repositories.

Refs #36

## Verification
- `bun run typecheck`
- `bun run lint:check` and `bun run format:check`
- `bun run build`
- real local Wrangler HTTPS Worker + D1 migration/start-flow/authorize redirect/PKCE flow; one pending row observed, then temporary state removed
- `bunx wrangler deploy --dry-run --config workers/identity/wrangler.toml`
- `bunx wrangler pages functions build functions --outdir /tmp/slop-pages-functions-build --project-directory .`
- workflow YAML parsed successfully

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -283,7 +283,7 @@ jobs:
               ;;
           esac
 
-      - name: Deploy to Cloudflare Pages
+      - name: Deploy private identity and Cloudflare Pages
         id: pages-deploy
         env:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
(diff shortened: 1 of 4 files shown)


lalalune (approved):
Reviewed the runtime-binding and trusted-release changes at 8cd23fdb2a6ed04548ad50e626c7270774bc815e. Local verification passed: projects/evaluations/cycles checks, typecheck, format, lint, 337 Vitest tests, 50 skill tests, production build, and 36 Chromium E2E cases across desktop/tablet/mobile. The configuration preserves private R2/D1/service bindings and the deploy remains restricted to the protected develop release job.

lalalune (changes requested):
Blocking production findings from a full runtime/security review:

1. **P1 — Pages runtime can deploy green while the API is dead.** The workflow proves the four identity Worker secret names but never verifies Pages `TRACE_AUTH_SECRET` / `OPERATOR_GITHUB_IDS`, the trace schema, the private R2 binding, or a live `api.slop.cash` response. `functions/api/v1/[[path]].ts` currently calls `OPERATOR_GITHUB_IDS.split(...)`; an absent value makes every API request 500. Add required Pages runtime inventory/migration gates and a post-deploy probe whose expected unauthenticated result is 401/403, never 5xx.

2. **P1 — preview deployments inherit production private storage.** Cloudflare Pages treats the top-level Wrangler config as shared configuration. These top-level D1/R2/service bindings therefore give preview Functions access to `slop-private`, permanent raw traces, and the production identity service. Define a production-only boundary and ensure previews receive no private production bindings (or isolated preview resources), with a contract test.

3. **P2 — identity release provenance is incomplete.** The Worker deploy is not tagged/message-bound to `GITHUB_SHA`, its deployment/version is not captured or verified, and a cancellation/Pages failure can leave an unrecorded mixed release. Bind and verify the Worker version to the exact release SHA before considering the Pages release complete.

4. **P2 — valid GitHub App client IDs are rejected.** `/^Iv[0-9A-Za-z]{18}$/` forbids the dot in GitHub's documented example `Iv1.8a61f9b3a7aba766`. Validate presence/equality or support GitHub's documented format instead of narrowing it incorrectly.

Focused verification on current head otherwise passed: deployment-contract 7/7, Pages Functions compilation, identity Worker dry-run with D1, and the D1/R2/service syntax.

PR #62, opened 2026-08-15
fix: fail closed on identity secret drift

NubsCarson:
## Summary
- document exact independent 32-byte base64url generation for both identity crypto keys
- normalize non-API exceptions at the Worker boundary so DOMException numeric codes cannot escape in the public JSON contract
- fail closed before persisting an OAuth flow when encryption key material is malformed
- lock zero additional GitHub App account permissions and key-generation commands into the deployment contract

This follows merged #61. The initially uploaded, unused malformed keys were replaced in Cloudflare before any Worker deployment. Payouts remain disabled.

Refs #36

## Verification
- `bun run typecheck`
- `bun run lint:check` and `bun run format:check`
- `bun run build`
- real local Wrangler HTTPS Worker + D1 migration/start-flow/authorize redirect/PKCE flow; one pending row observed, then temporary state removed
- identity Worker dry-run and deployment workflow YAML parse

### DIFF
--- a/scripts/deployment-contract.test.mjs
+++ b/scripts/deployment-contract.test.mjs
@@ -25,6 +25,10 @@ const identityWranglerConfiguration = readFileSync(
   join(packageRoot, "workers", "identity", "wrangler.toml"),
   "utf8",
 );
+const identityDeploymentGuide = readFileSync(
+  join(packageRoot, "workers", "identity", "README.md"),
+  "utf8",
+);
 const pagesHeaders = readFileSync(
   join(packageRoot, "public", "_headers"),
   "utf8",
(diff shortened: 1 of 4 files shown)


lalalune:
I recovered and independently verified the post-#61 hardening. I added one narrow follow-up commit to this canonical PR: the catch boundary now also normalizes non-Error throws instead of dereferencing `null`/primitives and dropping the response entirely. The added regression test proves a safe 500 and zero persisted flows. Focused identity/deployment tests, typecheck, format, and lint are green.

lalalune (approved):
Reviewed and verified at d6269c7582c9993c95c338fcfb363e01620ff6c0. The zero-permission App/key-generation contract is correct; malformed configuration and arbitrary non-Error failures now remain behind a sanitized 500 boundary without persistence. Local verification: 339 Vitest tests, 50 skill tests, typecheck, format, lint, production build, focused identity/deployment tests, and identity Worker dry-run.

PR #63, opened 2026-08-15
fix: fail closed on identity secret drift

lalalune:
## Summary
- preserve the post-#61 identity hardening commit that landed on the closed source branch
- require a zero-permission GitHub App and document independent generated 32-byte identity secrets
- normalize malformed error status/code values and all non-Error throws to a safe 500 response
- prove malformed encryption configuration and unknown failures cannot persist an OAuth flow or escape the response boundary

Refs #36

## Verification
- `bun run typecheck`
- `bun run format:check`
- `bun run lint:check`
- focused identity/deployment tests (14 passed after the additional unknown-throw case)
- `bun run build`
- `bunx wrangler deploy --dry-run --config workers/identity/wrangler.toml`

No production, DNS, wallet, payout, signing, or settlement action is included.

### DIFF
--- a/scripts/deployment-contract.test.mjs
+++ b/scripts/deployment-contract.test.mjs
@@ -25,6 +25,10 @@ const identityWranglerConfiguration = readFileSync(
   join(packageRoot, "workers", "identity", "wrangler.toml"),
   "utf8",
 );
+const identityDeploymentGuide = readFileSync(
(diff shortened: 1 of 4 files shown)


lalalune:
Closing as a duplicate of #62. The additional unknown-throw response-boundary hardening from this branch was cherry-picked onto #62 so no work is lost.

PR #64, opened 2026-08-15
fix: rate limit identity OAuth endpoints

NubsCarson:
Closes the remaining public-abuse and callback-log gate in #36.

- adds independent Worker-native limits for OAuth start and poll
- fails closed with stable 429/503 responses
- disables persisted invocation URLs so callback codes are not retained in Worker logs
- keeps counters outside D1 and preserves the two-second poll contract

Verification:
- full suite: 346 Vitest + 50 skill tests
- typecheck, Biome, production build, Worker dry-run
- real local Wrangler/D1 runtime: 12 starts returned 201 then 429; 120 polls returned 400 then 429

Payouts remain disabled and token volume remains non-financial.

### DIFF
--- a/scripts/deployment-contract.test.mjs
+++ b/scripts/deployment-contract.test.mjs
@@ -200,6 +200,23 @@ describe("slop.cash deployment contract", () => {
     expect(identityWranglerConfiguration).toContain(
       'database_id = "1b453124-2709-45af-8389-151a8105c461"',
     );
+    expect(identityWranglerConfiguration).toContain(
+      'name = "IDENTITY_START_LIMITER"',
+    );
+    expect(identityWranglerConfiguration).toContain('namespace_id = "81001"');
+    expect(identityWranglerConfiguration).toContain("limit = 12");
+    expect(identityWranglerConfiguration).toContain(
+      'name = "IDENTITY_POLL_LIMITER"',
+    );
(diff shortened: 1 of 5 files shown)


lalalune:
Independent review found one response-contract gap and patched it on this branch: the early 429/503 limiter responses now carry `Cache-Control: no-store` and `Referrer-Policy: no-referrer`, matching the identity handler's private-response boundary, with regression assertions for both paths.

The Worker binding shape, 60-second period, unique namespaces, and `invocation_logs = false` configuration match current Cloudflare primary documentation:
- https://developers.cloudflare.com/workers/runtime-apis/bindings/rate-limit/
- https://developers.cloudflare.com/workers/observability/logs/workers-logs/#invocation-logs

Local verification at the updated tree: typecheck, format, lint, 347 Vitest tests, 50 skill tests, and Wrangler dry-run showing both rate-limit bindings (12/60s and 120/60s). The existing start-by-IP choice is appropriate before authentication; the published per-location/eventual-consistency caveat remains correctly documented.

lalalune (approved):
Reviewed and verified at 2b42bce. The start/poll limits are path-scoped, use independent Cloudflare namespaces, fail closed on limiter errors, preserve the internal assertion route, and disable persisted invocation URLs. The added no-store/referrer policy closes the early-response privacy gap. Local typecheck, format, lint, 347 Vitest tests, 50 skill tests, and Worker dry-run are green.

PR #98, opened 2026-08-15
fix: preserve hidden Pages artifacts

lalalune:
Closes the production artifact gap that omitted `.well-known` endpoints from the trusted Pages bundle.

- include hidden files in the exact tested artifact
- verify the downloaded artifact against its manifest before any Cloudflare mutation
- fail closed when a hidden endpoint is missing
- pin the workflow contract and ordering

Validation: 357 Vitest tests; 69 skill tests; typecheck; lint; format; diff check.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -134,6 +134,7 @@ jobs:
         with:
           name: slop-pages-${{ github.run_id }}
           path: dist
+          include-hidden-files: true
           if-no-files-found: error
           retention-days: 7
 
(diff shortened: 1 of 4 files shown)


NubsCarson (approved):
Independent launch review at exact head f08acd1c1b60192a87700405ab08dcbd408211c4 on current develop ad4806f983b6a22e0a31bc1514f778f50e0363ca. The artifact uploader now preserves hidden `.well-known` endpoints, the deploy job verifies the downloaded bundle against its signed inventory before any Cloudflare mutation, and the deterministic fixture proves omission fails closed. The workflow contract pins ordering before scoped credentials. Hosted skill/data/unit/build/browser checks are green. This closes the reproduced manifest mismatch without weakening published-byte verification.

PR #107, opened 2026-08-15
fix: verify Cloudflare clean 404 route

lalalune:
## Summary
- verify the generated `404.html` bytes at Cloudflare Pages’ canonical `/404` route
- retain manual redirect handling for every other deployment artifact
- cover the clean-route mapping in the deterministic manifest tests

## Production evidence
The deployed exact artifact for run 31912241747 fails only because `/404.html` returns Cloudflare’s expected 308 to `/404`. With this change, the same downloaded artifact verifies all 152 public files and the manifest byte-for-byte at `https://slop.cash`.

## Verification
- `bunx vitest run scripts/dist-manifest.test.mjs scripts/deployment-contract.test.mjs` (14/14)
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `git diff --check`
- live exact-artifact verification: 152 files + manifest

### DIFF
--- a/scripts/dist-manifest.mjs
+++ b/scripts/dist-manifest.mjs
@@ -308,9 +308,15 @@ function remoteUrl(origin, path, verificationToken, attempt) {
     .split("/")
     .map((part) => encodeURIComponent(part))
     .join("/");
-  // Pages canonically serves the entry document at the apex and redirects
-  // /index.html, so its bytes must be checked at the public route.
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged PR #108, which implements the same clean-URL verification for all Cloudflare Pages HTML routes. Exact-head CI and live 152-file verification are green; no duplicate change is needed.

Issue #115, opened 2026-08-15
Add verified project creators and explicit IP policy to self-serve onboarding


lalalune:
## Outcome

Anyone can draft a project; activation requires GitHub identity, immutable repository authority proof, a verified project creator/steward, explicit payout operators, and truthful, versioned IP and prize terms. The UI presents one short creator/license line and one Terms link—no chips, repeated labels, or ownership marketing unsupported by signed documents.

## Authority model

Add strict schema fields for:

- `steward`: display name, individual/organization/DAO/collective kind, immutable GitHub actor ID/node ID/login/type/profile URL, optional website.
- `authority`: role, immutable repository ID, proof URL, commit SHA, file digest, verified timestamp.
- private `projectOperators`: GitHub numeric actor ID plus owner/editor/settler roles; these do not imply copyright.

Activation requires a merged target-repository `.github/slop-project.json` at an immutable commit. It binds Slop project ID/repository, steward actor ID, authorized operators, project policy digest, and proposal. Repository transfer, proof removal, integration-branch change, or material terms drift pauses management and new runs. Historical receipts remain pinned.

## IP schema

Keep stewardship, payout authority, copyright, repository license, inbound contribution terms, and external-prize rules separate. Add:

- copyright model: sponsor-owned, contributor-retained, mixed, or unknown;
- exact legal copyright holder where claimed;
- SPDX license plus immutable LICENSE URL/commit/digest;
- inbound mode: license, CLA, assignment, DCO, or mixed, with immutable terms/version/digest and acceptance mechanism;
- assignment assignee/instrument/version/digest when required;
- external organizer, rules URL/capture/digest/version, eligibility, allocation authority, and separate IP terms.

A project payment never transfers IP unless a separate valid signed assignment says so. Material policy changes pause new work and require a new acknowledgement; never apply terms retroactively.

## Confirmed project intent

- **Proximity Prize:** prize proceeds are split among the named contributors according to the final author-approved allocation; organizer rules remain controlling. This is payout allocation, not a copyright claim.
- **ASI:** MIT is intended, governed for `$SLOP` holders. Public copyright ownership must name an exact legal-capable DAO wrapper/foundation/association and bind a governance resolution plus signed assignments; literal shifting token holders are not a workable title holder. Until supplied, show the DAO as steward/governance beneficiary and copyright as unknown/mixed.
[...]

Issue #114, opened 2026-08-15
Build direct project funding and honor-system payout MVP


lalalune:
## Outcome

Any verified GitHub user can fund a project directly, without Slop receiving or controlling money. Each project publishes its own receiving addresses. Donors send funds from their own wallets to the project-controlled address and may attach a public transaction reference for attribution. Project creators distribute rewards from their own wallets and publish payout evidence.

Slop is a public record and coordination layer, not a custodian, exchange, escrow, payment processor, or wallet provider.

## Fixed boundaries

- Slop never generates, receives, stores, exports, or signs with a private key.
- Slop never controls a project wallet, donor wallet, contributor wallet, or funds in transit.
- No custody provider, smart contract, escrow, bridge, swap, multisig, embedded wallet, or platform withdrawal flow.
- Never imply that a GitHub login proves wallet ownership.
- Use **Fund this project**, not tax-deductible donation language.
- A reported transfer is not called verified or paid unless its public chain evidence passes the applicable deterministic verifier.

## Project funding configuration

The reviewed project manifest may publish one receiving address per supported network:

- Solana mainnet: USDC first;
- Base: USDC;
- Ethereum mainnet: USDC;
- Bitcoin mainnet: BTC.

Every address change is a reviewed manifest change with a visible effective time. Historical funding records remain bound to the address and manifest revision that were active when recorded. Slop does not infer a project from an arbitrary transfer.

## Donor flow

1. Open **Fund this project**.
2. Choose a supported network and see the exact asset, address, copy action, QR code, and explorer link.
3. Send directly from the donor's wallet to the project-controlled address.
4. Optionally submit the transaction ID and choose public GitHub attribution or anonymous display.
5. Slop publishes one of three explicit states:
   - **Self-reported** — submitted by the donor or project, not independently confirmed;
   - **Verified on-chain** — exact recipient, asset, amount, success, and required finality were independently checked;
   - **Disputed** — evidence conflicts, was replaced by a reorg, or no longer satisfies verification.

The donor profile shows only publicly attributed records and separates self-reported from verified totals. Project totals never combine the two into one number.

## Git-backed MVP records

- `funding/<project>/<chain>/<transaction-id>.json` is append-only reviewed public evidence.
- Records bind project manifest revision, network, asset, transaction ID, recipient, integer amount, observed finality, donor attribution choice, and verifier version.
[...]

PR #138, opened 2026-08-16
fix: trust project transition enforcement

lalalune:
## Summary

- run project-policy transition validation from an immutable, base-controlled `pull_request_target` workflow with read-only contents access, no secrets or dependency install, and the PR head fetched only as bounded Git object data
- compare immutable base/head manifest trees without checking out or executing PR-controlled code, while preserving a post-merge develop push gate
- enforce append-only funding route history: immutable prior fields, at most one successor, and an exact non-overlapping close/start rotation
- add workflow-contract and adversarial funding-history tests

Addresses #115 and #114. This is a narrow follow-up to the post-merge audit of #136; it does not close either product issue.

## Security model

The trusted job checks out `github.event.pull_request.base.sha` and verifies that exact checkout before fetching `refs/pull/<number>/head`. It verifies the fetched object equals `github.event.pull_request.head.sha`, then executes the checker and schema modules from the trusted base only. It never checks out or invokes the PR head, installs dependencies, accesses secrets, or grants write permission.

Schema/checker evolution therefore fails closed: a PR cannot weaken the checker that evaluates itself. The develop push workflow separately compares immutable `before` and current SHAs.

## Validation

- `bun install --frozen-lockfile --ignore-scripts`
- `bun run projects:check`
- `bun run evaluations:check`
- `bun run cycles:check`
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- focused transition/workflow suite: 17/17 passing
- isolated `scripts/prepare-site.test.ts`: 9/9 passing
- `bun run build`
- exact Git-object transition check from `f6ad561395edbb2c8376c0c3f468249d5c80dee2` to `00f3d2a9970d6906086cc6287f9ecd1f76f94d26`: 4/4 projects validated

A parallel full Vitest run was attempted twice. The first reached 435 passing tests with only the 10-second `prepare-site` setup timeout; the second, under host load, reached 430 passing and timed out in the same setup plus five existing 5/15-second UI/installer tests. The affected prepare-site suite passes 9/9 in isolation; the targeted UI/installer retry remained timing-sensitive under the same host load. No product assertion failure remains in the changed transition tests.

## Required ruleset follow-up (not changed here)

Ruleset `20220584` is active with no bypass actors, but currently has no required status checks. After this workflow exists on `develop`, require the status context **Trusted project transition gate** with `strict_required_status_checks_policy: true` (or the equivalent current-branch requirement), retaining zero bypass actors. This closes the base-advance gap; `pull_request_target` alone does not rerun merely because `develop` advances.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -90,11 +90,12 @@ jobs:
 
       - name: Validate project registry and every skill
         env:
-          PROJECT_POLICY_BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
+          PROJECT_POLICY_BASE_SHA: ${{ github.event.before }}
+          PROJECT_POLICY_HEAD_SHA: ${{ github.sha }}
         run: |
           set -euo pipefail
-          if [ "${{ github.event_name }}" = "pull_request" ] || [ "${{ github.event_name }}" = "push" ]; then
-            node scripts/check-project-transitions.mjs "$PROJECT_POLICY_BASE_SHA"
+          if [ "${{ github.event_name }}" = "push" ]; then
+            node scripts/check-project-transitions.mjs "$PROJECT_POLICY_BASE_SHA" "$PROJECT_POLICY_HEAD_SHA"
           fi
           bun run projects:check
           bun run evaluations:check
--- a/.github/workflows/project-transitions.yml
+++ b/.github/workflows/project-transitions.yml
@@ -1,0 +1,44 @@
+name: Trusted project transition gate
+
+on:
+  pull_request_target:
+    branches: [develop]
+    types: [opened, synchronize, reopened, ready_for_review]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: trusted-project-transition-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  project-transitions:
+    name: Trusted project transition gate
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    steps:
+      # Execute only the checker committed at the immutable base SHA. The PR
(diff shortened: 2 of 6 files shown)


lalalune (commented):
Careful review with local verification from the PR head in a clean worktree:

- `bun test scripts/deployment-contract.test.mjs`: 9/9 (169 expectations)
- `vitest src/lib/project-policy.test.ts`: all passing, including the new adversarial funding-history cases (alteration, deletion, reorder, backfill, close-without-successor, wrong cutover, multi-append, pre-closed append)
- `node scripts/check-project-transitions.mjs <develop-sha> <head-sha>`: 4/4 projects validated

Security model checks out: the `pull_request_target` job executes only base-committed code from a full-SHA-pinned checkout, `contents: read`, no secrets, no dependency install, `persist-credentials: false`, asserts the checked-out HEAD equals the event base SHA, and fetches the PR head strictly as git object data with an exact-SHA equality gate. A PR cannot weaken the checker that evaluates it, and the develop push gate remains as the merge-time backstop.

Follow-up after merge: ruleset 20220584 should require the "Trusted project transition gate" status context to close the stale-base window.

lalalune:
Exact-head CI is green for `00f3d2a9970d6906086cc6287f9ecd1f76f94d26` against unchanged `develop` base `f6ad561395edbb2c8376c0c3f468249d5c80dee2`.

Run: https://github.com/elizaOS/slopdotcash/actions/runs/31964342623

The 7m52s quality job passed frozen install, dependency audit, project/skill validation, live ledger generation, cycle verification, typecheck, format/lint, the complete unit suite, production build, and desktop/mobile Chromium E2E. Production deployment correctly skipped for this PR. No merge was performed.

PR #166, opened 2026-08-18
fix(skills): detect direct terms-preflight runs through symlinked paths

PrototypicNoise:
## Bug

The direct-invocation gate in every project's `terms-preflight.mjs` compares `import.meta.url` against `new URL(\`file://\${process.argv[1]}\`)`. Node resolves the entry module to its **realpath**, so when the CLI is invoked through a **symlinked path** — exactly the layout produced by versioned skill installs, where `skills/<name>` is a symlink into a version store — the two sides diverge, `direct` is false, and the script **silently exits 0 without performing the preflight**. An operator or wrapper reading exit 0 as "terms verified" has verified nothing. (Found in the field: our agent's first preflight run "passed" this way; the receipt CLI's imported preflight then failed honestly.)

Repro on this branch's base:
```
$ ln -s .../skills/contribute-to-eliza/scripts/terms-preflight.mjs /tmp/l/tp.mjs
$ node /tmp/l/tp.mjs --project eliza --authority file:///tmp/   # forbidden flag
$ echo $?
0        # silent no-op — should be a loud refusal
```

## Fix

Align all four project copies (asi, eliza, delta-star, heir-elements-sdk) with the realpath comparison **already used by `run-receipt.mjs` and `live-report.mjs`** in this repo:

```js
const direct =
  typeof process.argv[1] === "string" &&
  existsSync(process.argv[1]) &&
  realpathSync(fileURLToPath(import.meta.url)) ===
    realpathSync(process.argv[1]);
```

This also fixes the same comparison's latent breakage on paths containing spaces/percent-encodable characters (`file://` string-building vs `pathToFileURL` semantics).

## Test (failing-test-first)

`skill-tests/terms-preflight.test.ts` gains a symlink regression that spawns **node explicitly** — node resolves the entry realpath; bun preserves the symlinked path, so the repo's bun runner cannot reproduce the divergence with `process.execPath` (comment in-test documents this). On base: 4 pass / **1 fail**; with the fix: **5 pass**.

## Verification

Commit: `15e49361f3888027c1eaa66a490d9a02b10e33ff`
```
bun test ./skill-tests            # 80 pass, 0 fail
biome check (5 touched files)     # clean
```

## Disclosure

Client `claude-code`, provider `anthropic`, model `claude-fable-5` (agent-authored, human-operated).

---
🤖 Built with [Emblem:build](https://emblem.build)

### DIFF
--- a/skill-tests/terms-preflight.test.ts
+++ b/skill-tests/terms-preflight.test.ts
@@ -1,5 +1,8 @@
 import { spawnSync } from "node:child_process";
 import { createHash } from "node:crypto";
+import { mkdtempSync, rmSync, symlinkSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import { fileURLToPath } from "node:url";
 import { afterEach, describe, expect, it } from "vitest";
 import { preflight } from "../skills/contribute-to-eliza/scripts/terms-preflight.mjs";
@@ -139,5 +142,32 @@ describe("project terms preflight", () => {
     );
     expect(direct.status).toBe(1);
     expect(direct.stderr).toMatch(/authority overrides are forbidden/u);
+  });
+
+  it("detects direct invocation through a symlinked script path", () => {
+    const script = fileURLToPath(
+      new URL(
+        "../skills/contribute-to-eliza/scripts/terms-preflight.mjs",
+        import.meta.url,
+      ),
+    );
+    const linkRoot = mkdtempSync(join(tmpdir(), "terms-preflight-link-"));
+    const link = join(linkRoot, "terms-preflight.mjs");
+    try {
+      symlinkSync(script, link);
+      // Spawn node explicitly: node resolves the entry module to its realpath,
+      // so import.meta.url and a symlinked argv[1] diverge — the case that made
+      // the old comparison silently classify a direct CLI run as an import.
+      // (bun preserves the symlinked path, so bun cannot reproduce the bug.)
+      const viaSymlink = spawnSync(
+        "node",
+        [link, "--project", "eliza", "--authority", "file:///tmp/"],
+        { encoding: "utf8" },
+      );
+      expect(viaSymlink.status).toBe(1);
+      expect(viaSymlink.stderr).toMatch(/authority overrides are forbidden/u);
+    } finally {
+      rmSync(linkRoot, { recursive: true, force: true });
+    }
   });
 });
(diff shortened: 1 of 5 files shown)


lalalune (approved):
Exact-head review at `15e49361f3888027c1eaa66a490d9a02b10e33ff`: the prior `file://${process.argv[1]}` comparison can silently misclassify Node entrypoints reached through installed-skill symlinks. The realpath comparison matches the existing receipt/live-report pattern and preserves import behavior. Verified `git diff --check`, the focused 5-case terms-preflight suite, all 80 skill tests, and repository formatting.

wbaxterh (approved):
**Reviewed at head `15e4936` — bug reproduced at the merge-base, fix verified live, approving.**

**What I ran** (Linux, node v24.15.0, fresh checkout):

1. **Repro at merge-base `527d111`:** symlinked the eliza copy of `terms-preflight.mjs` into a temp dir and invoked it with a forbidden `--authority file:///tmp/`. Exit **0**, empty stderr: the preflight silently no-ops through a symlink, which is the worst failure shape for a gate whose only job is to refuse. Mechanism confirmed: node resolves the entry module to its realpath for `import.meta.url` while `argv[1]` keeps the symlink path, so the old equality could never hold under a versioned-install layout.
2. **Same invocation at head:** exit **1**, `direct CLI authority overrides are forbidden`. Direct non-symlinked invocation still detected as well.
3. **Copy audit:** exactly four vendored `terms-preflight.mjs` copies exist in the tree, and all four are byte-identical after the fix (single md5). No copy was missed.
4. **`bun test skill-tests/terms-preflight.test.ts`:** 5/5 pass, including the new symlink case.

Two details worth calling out as strengths: the `existsSync` guard is load-bearing, not decoration (`realpathSync` throws `ENOENT` on a missing `argv[1]`, verified), and the regression test spawns real `node` with the correct rationale in-comment (bun preserves the symlinked `argv[1]`, so bun cannot reproduce the bug) while asserting the enforcement side effect rather than just the exit status.

**Non-blocking follow-up suggestion:** the four vendored copies being byte-identical is currently maintained by hand. A cheap CI assertion that all `skills/*/scripts/terms-preflight.mjs` share one hash would turn a future partial fix from silent drift into a red build. Happy to file it as an issue if wanted.

— [claude-code-wbaxterh]

lalalune (approved):
Re-approved exact updated head `0a31b3d66c83e179ef9106a1af463a333a222eb7` after synchronizing current `develop`. The five-case terms-preflight suite passes, including direct invocation through a symlinked script path, and the merged evidence-head verifier self-test passes all 15 cases. The combined diff remains scoped and fail-closed.

PR #165, opened 2026-08-18
Ship outcome-weighted scoring v2

lalalune:
## Summary

- replace flat/capped activity scoring with maintainer-ratified integer-third work units, automated review credit, and fixed trace/usage evidence bonuses effective August 2026
- keep every reviewed project open to contributions even when repository authority, license, inbound terms, or the legacy manifest status are unresolved; unknowns are disclosed and signed instead of blocking runs
- allow nullable policy digests in v2 receipts and preserve exact historical policy bindings when declared terms later become available
- publish the scoring protocol, truthful funding state, and a schema bridge so pull-request browser checks can consume the currently deployed ledger safely

## Validation

- exact head: 9c12d72
- rebased onto current develop: 86d1d4c
- trusted project transition gate: 4 projects validated
- focused policy/schema/receipt/UI tests: 78 passed; terms-preflight CLI tests: 4 passed
- exact-head real-browser matrix: 36 responsive and accessibility tests passed; Pages artifact contract: 1 passed
- live GitHub snapshot: schema 6 / slop-score-v2, 103 leaders, 4,203 score events
- full verifier passed locally before final rebase; protected CI is the final aggregate verification

## Boundaries

- any declared model/provider/client may contribute; model choice and raw token volume never change score
- unknown legal/authority facts are not invented
- payouts remain disabled until independent funding and settlement evidence exists

Closes #164

PR #170, opened 2026-08-18
Separate contribution access from authority evidence

lalalune:
Stages the project-policy compatibility needed before activating reviewed projects with disclosed unknown authority or terms. This PR changes no project status, receipt cutover, funding state, or payout state. Verified projects still require an immutable proof-bound receipt cutover; funding and payouts remain independently evidence-gated.\n\nValidation: 20 focused schema/policy tests, typecheck, lint, format, AGENTS/CLAUDE identity, and diff check pass.\n\nThis must merge before #165 so the trusted base checker can validate the all-four activation without a protection bypass.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -77,9 +77,11 @@ skills/contribute-to-<project-id>/
 skills/review-<project-id>-contributions/
 ```
 
-New projects begin paused. Verify immutable repository and actor IDs, repository
-license facts, GitHub stewardship, integration branch, reward policy, and
-failure paths before activation. Stewardship is a GitHub identity only.
+Once reviewed and merged, new projects accept contributions. Verify immutable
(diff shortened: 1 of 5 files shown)


lalalune:
Closing after exact-head review at `b1f050b4f27ecfe4454a3c3449b4881eef48c819`. This is not a compatibility-only refactor: it deletes the fail-closed guarantees that an active project has verified repository authority and known mandatory terms. That directly conflicts with the repository contract that new projects begin paused until immutable repository/actor IDs, license facts, stewardship, integration branch, reward policy, and failure paths are verified. It would also make an active project with unverified authority and unknown license/inbound terms schema-valid solely to unblock #165. Contribution access can be made broad without representing an unverified project as active; model/client eligibility and project activation are separate boundaries.

lalalune:
Closed rather than weakening project activation authority to make a downstream branch pass.

PR #176, opened 2026-08-18
Revert unsafe scoring and authority bundle

lalalune:
## Summary

- revert #173’s rejected scoring-v2, relaxed authority, workflow, receipt, terms-preflight, and UI changes while preserving the independently reviewed #174 queue-ordering work
- preserve the immutable, proof-bound ASI and Eliza receipt-policy activations introduced on `develop`; append-only policy history is not rolled back
- retain fail-closed activation semantics for projects without verified authority and mandatory terms
- remove the unrelated `Closes #164`; that wallet issue remains unverified and its public actor claim endpoint returns 404

## Validation

- immutable transition checker passes from base `55022ff5c77d286c724581a549c02b70687a45e4`
- `bun install --frozen-lockfile`
- `bun run verify`
- 59 Vitest files / 567 tests passed
- 81 skill tests passed
- production build and 170-file deployment manifest passed
- `AGENTS.md` and `CLAUDE.md` remain byte-identical

Exact validated head: `033ef6fe569d071b602730870138475970e5e436`.

PR #177, opened 2026-08-18
Bind score bonuses to verified outcomes

lalalune:
## Outcome

Keeps score v2 and closes the integrity gaps identified in #176 without reverting the requested payout policy.

## Security and accounting changes

- verifies Ed25519 run receipts before automated review credit, maintainer proposal acceptance, or evidence bonuses
- binds each 15% trace and 10% exact/bounded-usage bonus to the exact actor and PR/comment outcome
- applies payout weight per outcome, so one receipt cannot boost unrelated work by the same actor
- keeps public leaderboard score and project score identical; evidence changes payout weight only
- rejects published bonus metadata without the matching exact receipt
- recognizes both GitHub review objects and review comments in the causal attribution invariant

## Policy retained

- August 1 score-v2 migration
- integer thirds with rounding only after aggregation
- no account caps or diminishing post-August credit
- maintainer-ratified effort tiers and co-review for XL/security cases
- automated reviewer and maintainer compute in the same project pool
- $2 minimum payout accrual
- all four registered projects accepting contribution runs
- no finite 60-second test ceiling

## Validation

- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- focused scoring/project tests: 93 passed
- Vitest full suite: 61 files, 578 passed in 211.78s
- exact isolated Bun ownership test: 1 passed in 14.87s

The aggregate Bun runner killed that same child process during cleanup (`status: null`); the exact owning test passes alone. No finite timeout is configured.

This supersedes the wholesale revert in #176.

### DIFF
--- a/scripts/generate-leaderboard.ts
+++ b/scripts/generate-leaderboard.ts
@@ -2855,20 +2855,8 @@ export async function generateLeaderboardFromGitHub(
     verificationWindowFrom: verificationWindowFrom.toISOString(),
     verifiedEvidence: evidenceVerification.artifacts,
     evaluatedContributions: options.evaluatedContributions ?? [],
+    verifyRunReceipt: verifyRunReceiptSignature,
   });
-  for (const attribution of snapshot.attributions) {
-    if (attribution.run === null) continue;
-    try {
-      verifyRunReceiptSignature(attribution.run);
-    } catch (error: unknown) {
-      snapshot.invalidAttributionMarkers.push({
-        sourceId: attribution.sourceId,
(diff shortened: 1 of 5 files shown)


lalalune:
Post-merge audit found one remaining exact-outcome boundary: a valid signed receipt from a different registered repository could still receive the contribution evidence bonus because the join checked actor/artifact but not `receipt.repositoryId === event.repository`. PR #179 fixes that narrowly and adds an ASI-receipt-on-Eliza-PR adversarial regression.

PR #179, opened 2026-08-18
Bind run receipts to GitHub artifacts

lalalune:
## Security outcome

Closes the remaining cross-repository replay gap after #177: a valid signed receipt for one registered repository can no longer contribute payout weight or model attribution when copied onto an outcome in another repository.

## Changes

- attach immutable artifact URLs to pull request bodies and issue text sources
- require every verified run receipt repository to match its GitHub artifact repository
- suppress visible-declaration fallback for a repository-mismatched machine receipt
- add an adversarial ASI-receipt-on-Eliza-PR regression proving zero evidence bonus and no accepted attribution

## Validation

- exact head: `9e317902e40a38b7ca5eadacec566fab77c71a23`
- leaderboard tests: 84 passed
- full Vitest: 61 files / 579 passed
- Bun skill aggregate: 80 passed; the known process-cleanup case was killed with `status: null`
- exact owning Bun test rerun alone: 1 passed in 13.67s
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`

### DIFF
--- a/src/lib/leaderboard.test.ts
+++ b/src/lib/leaderboard.test.ts
@@ -567,6 +567,37 @@ describe("score v2 work units", () => {
       "has no exact receipt for its evidence bonus",
     );
   });
+
+  it("rejects a signed receipt replayed onto an outcome in another repository", () => {
+    const body = reviewAttribution({
+      projectId: "asi",
+      repositoryId: "elizaOS/asi",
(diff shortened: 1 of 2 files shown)


lalalune:
Rebased onto current `develop` after #178 merged. Exact current head: `1a58f7ce108b836876d57a64e7758c68e449ac06`; focused leaderboard suite passes 84/84 at this head.

PR #172, opened 2026-08-18
Ship outcome-weighted scoring v2

lalalune:
Supersedes closed #165 at the exact rebased head.\n\n- replaces flat/capped activity scoring with maintainer-ratified integer-third work units, automated review credit, and fixed trace/usage evidence bonuses effective August 2026\n- groups split work, removes account caps, accrues awards below USD 2, and includes maintainer/reviewer compute in the same project pool\n- makes all four existing projects accept contribution runs and bounded receipt acknowledgements while explicitly preserving unknown authority or terms\n- migrates the real deployed schema-5 ledger for untrusted PR browser checks instead of fabricating empty data\n- removes every 60-second test ceiling\n\nValidation on the rebased head: immutable four-project transition passes; project/evaluation/funding/cycle/protocol/dependency/type/format/lint gates pass; 61 Vitest files and 575 tests pass in 310 seconds; the loaded aggregate run had one externally killed Bun fixture subprocess, and the exact case passed independently in 5.6 seconds; deployed-ledger schema bridge and prior exact-data desktop/mobile Playwright validation pass.\n\nMonthly payout drafting remains evidence-gated until reviewed immutable funding instruments prove committed funds. Delta Star remains an external-prize allocation rather than platform money.\n\nCloses #164.

PR #180, opened 2026-08-18
fix: exclude post-merge slop-reviews instead of failing the snapshot

wbaxterh:
## Summary

- Exclude a `slop-review` posted after merge instead of throwing from `createLeaderboardSnapshot`.
- Add the regression: a post-merge automated review scores nothing and the snapshot still generates (the merge itself still scores its 1/3).

## Why now

The v2 review loop landed on `develop` tonight with this path intact: a structurally valid `slop-review` comment (well-formed record + receipt join) posted after merge hits `throw new TypeError("slop-review was posted after merge")` inside the source loop, while every other disqualifying condition in that loop `continue`s. Comment authors are unprivileged, so one late comment on any detailed merged PR fails every subsequent trusted ledger regeneration, including the deploy path. `AGENTS.md` already defines the intended contract ("Exclude bots, self-review, post-merge review, and repeated low-value comments"); this makes the code match it.

This is blocker 1 from my review on #172, now applying to `develop` directly. Blockers 3 (self-ratification guard) and 4 (sub-2-USDC carry for cycle-inactive actors) are coming as separate scoped PRs.

## Validation

- New regression fails on unfixed `develop` with the exact production `TypeError` at `leaderboard.ts:3120`, passes with the fix.
- `bun test src/lib/leaderboard.test.ts` — 84 passed, 0 failed.
- `bun run lint:check`, `bun run format:check` — clean.
- `bun run projects:check` — registry current.
- Not run here: `bun run test:e2e` and `bun run build` (no UI or build-path change; scoring library + its unit test only). Reporting the boundary precisely per AGENTS.md.

— [claude-code-wbaxterh]

### DIFF
--- a/src/lib/leaderboard.test.ts
+++ b/src/lib/leaderboard.test.ts
@@ -567,6 +567,56 @@ describe("score v2 work units", () => {
       "has no exact receipt for its evidence bonus",
     );
   });
+
+  it("excludes a slop-review posted after merge without failing the snapshot", () => {
+    const reviewer = actor("reviewer");
+    const body = reviewAttribution(
+      {
+        provider: "openai",
+        model: "gpt-5.6-sol",
+        client: "codex",
+      },
+      {
+        schemaVersion: "2",
+        splitRisk: "none",
+        effortBand: "small",
+        complexity: "moderate",
+        impact: "meaningful",
+        reviewLoad: "standard",
+        recommendedTier: "small",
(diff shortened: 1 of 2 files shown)


lalalune:
The finding and patch are valid, but this head conflicts with merged #179 and would drop its receipt-to-artifact protection. Superseded by #184, which preserves your authorship and exact regression on current `develop` (85/85 focused tests).

PR #181, opened 2026-08-18
fix: require a second maintainer for self slop-score ratification

wbaxterh:
## Summary

- Require a second-maintainer co-ratifier for any `slop-score` whose author is the pull request's own author, regardless of tier.
- Add the regression pair: a self slop-score without a co-ratifier fails snapshot generation with a precise error; the same record with an independent maintainer's `slop-score-approval` scores normally.

## Why

`assertScoreRatificationContext` validates project, node, head SHA, maintainer association, and immutability, but never compares the ratification author to the PR author, and the merge-score path applies `record.scoreThirds` unconditionally. The co-ratifier requirement only fires at XL/exceptional or `securityRisk !== "none"`, so a lone maintainer could ratify their own PR to `large` (8 points) with a self-produced proposal. `protocol/scoring-v2.md` already states related-party cases require a second maintainer; this makes the code enforce it uniformly. The existing co-ratifier validation (must differ from the ratification author, immutable maintainer comment, matching approval record) then guarantees the second party is independent.

This is blocker 3 from my review on #172, applying to `develop` since tonight's slice merges. Blocker 1 is #180; blocker 4 (sub-2-USDC carry for cycle-inactive actors) comes next as its own scoped PR.

Design note: the guard throws, matching the loop's existing fail-closed convention for maintainer-authored records (supersede-chain violations throw the same way); unlike the post-merge case in #180, this input is privileged, so failing loudly is the correct shape.

## Validation

- New regression fails on unfixed `develop` (no throw occurs where one is required), passes with the guard.
- `bun test src/lib/leaderboard.test.ts` — 84 passed, 0 failed.
- `bun run lint:check` — clean.
- Not run here: `bun run test:e2e` and `bun run build` (scoring library + unit test only). Reporting the boundary precisely per AGENTS.md.

— [claude-code-wbaxterh]

### DIFF
--- a/src/lib/leaderboard.test.ts
+++ b/src/lib/leaderboard.test.ts
@@ -500,6 +500,99 @@ describe("score v2 work units", () => {
     ).toHaveLength(1);
   });
 
+  it("rejects a self slop-score without a second maintainer co-ratifier", () => {
+    const author = actor("author");
+    const pr = pullRequest({
+      id: "PR_SELF_SCORE",
+      number: 9,
+      createdAt: "2026-08-17T08:00:00.000Z",
+      updatedAt: "2026-08-18T03:00:00.000Z",
+      mergedAt: "2026-08-18T03:00:00.000Z",
+      author,
+    });
+    const record = {
+      schemaVersion: "1",
+      ruleVersion: "slop-score-v2",
+      projectId: "eliza",
+      pullRequestNodeId: pr.id,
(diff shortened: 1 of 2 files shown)


lalalune:
Valid finding and regression, now superseded by current-develop #184 with your authorship preserved. Combining the two adjacent leaderboard protections avoids a second full protected browser cycle.

PR #182, opened 2026-08-18
fix: carry prior below-minimum accrual for cycle-inactive actors

wbaxterh:
## Summary

- Emit allocation rows for actors who hold a prior accrued balance but have no events in the current cycle, so a quiet month never discards a sub-2-USDC carry.
- Include those balances in the manifest's `carriedMinor` and `suggestedMinor` totals.
- Add `priorActorLogins` to the proposal input (backward compatible; both existing callers unchanged) and fail closed when a carried balance's login cannot be resolved from the prior manifest.
- Regression: a below-minimum carry for a cycle-inactive actor produces a `held-below-minimum` row with score 0 and joins `carriedMinor`; a carry without a resolvable login throws.

## Why

`createRewardCycleProposal` built `carriedMinor` and allocations exclusively from `view.leaders`, i.e. actors with events in the current cycle. An actor holding a `held-below-minimum` accrual who had a quiet month got no allocation row, so their carry silently vanished from the next manifest. `cycles/README.md` promises sub-2-USDC awards "accrue without being discarded or redistributed"; this makes the proposal generator keep that promise. Accrual is a debt to the actor, not a reward for this cycle's activity.

This is blocker 4 from my review on #172, applying to `develop` since tonight's slice merges. Blocker 1 is #180 and blocker 3 is #181; this completes the set I offered to fix. Blocker 2 (review-credit authority) needs a design decision first, and I am happy to write that up as a short proposal if useful.

## Validation

- New regression fails on unfixed `develop` (the carried actor's row is absent), passes with the fix.
- `bun test src/lib/reward-cycle.test.ts src/lib/rewards.test.ts` — 19 passed, 0 failed.
- `bun run lint:check` — clean (206 files).
- Not run here: `bun run test:e2e` and `bun run build` (reward library + unit test only). Reporting the boundary precisely per AGENTS.md.

— [claude-code-wbaxterh]

### DIFF
--- a/src/lib/reward-cycle.test.ts
+++ b/src/lib/reward-cycle.test.ts
@@ -73,6 +73,46 @@ describe("reward cycle proposals", () => {
     });
   });
 
+  it("carries a prior below-minimum accrual for a cycle-inactive actor", () => {
+    const base = {
+      cycleId: "2026-07",
+      generatedAt: "2026-08-02T00:00:00.000Z",
+      projectId: "eliza" as const,
+      snapshot: closedSnapshot(),
+      sourceSnapshotSha256: SOURCE_SHA,
+      priorAccruedMinor: new Map([["U_quiet", "1500000"]]),
+    };
+    expect(() => createRewardCycleProposal(base)).toThrow("has no prior login");
+
+    const proposal = createRewardCycleProposal({
+      ...base,
+      priorActorLogins: new Map([["U_quiet", "quiet-contributor"]]),
+      wallets: new Map([
+        [
+          "U_quiet",
+          {
+            ...wallet(),
+            sourceUrl: `https://github.com/quiet-contributor/quiet-contributor/blob/${PROFILE_COMMIT}/README.md`,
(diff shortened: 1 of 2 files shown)


lalalune:
Closing because the regression proves only a library-injected fixture, not the production lifecycle. Three required paths are unchanged: (1) `scripts/prepare-reward-cycle.ts` never loads a prior approved allocation or passes `priorAccruedMinor` / `priorActorLogins`; (2) its bounded wallet lookup covers only current-cycle leaders, not carried-only actors; and (3) `scripts/sync-cycle-index.ts` regenerates the baseline without prior state, so it would reject the added carried row as an actor added outside the frozen snapshot. As submitted, the monthly automation still cannot create or validate the promised quiet-month carry. A safe successor must derive carry from immutable prior allocation state, feed the same maps into preparation and validation, and test the actual monthly command across two cycles. No cycles exist yet, so closing this incomplete financial change preserves the fail-closed launch boundary.

PR #184, opened 2026-08-18
Exclude post-merge reviews without failing snapshots

lalalune:
Supersedes #180 on current `develop` while preserving the cross-repository receipt binding from #179.

This keeps the original author and fix: unprivileged `slop-review` comments created after merge are excluded rather than throwing and denying all future snapshot generation.

Validation on the rebased exact head:
- leaderboard suite: 85/85 passed
- typecheck passed
- lint and format checks passed

Original analysis and attribution remain in #180.

### DIFF
--- a/src/lib/leaderboard.test.ts
+++ b/src/lib/leaderboard.test.ts
@@ -593,7 +593,7 @@ describe("score v2 work units", () => {
     ).toHaveLength(1);
   });
 
-  it("awards automated review credit only after receipt signature verification", () => {
+  it("awards automated review credit only after signature verification and maintainer ratification", () => {
     const reviewer = actor("reviewer");
     const body = reviewAttribution(
       {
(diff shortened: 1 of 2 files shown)


lalalune:
Stacked the independently valid #181 guard while preserving original authorship: a maintainer may no longer be the sole ratifier of their own outcome. Exact focused suite now passes 86/86 plus type/lint/format.

lalalune:
Closed the remaining review-authority gap identified in #182’s analysis: a signed `slop-review` now earns automated review credit only when the current immutable maintainer `slop-score` explicitly references that review node. The regression proves invalid-signature, valid-but-unratified, and valid-ratified states. Also corrected detailed-coverage normalization for automated-review and ratifier event IDs. Focused suite: 86/86; type/lint/format pass.

PR #186, opened 2026-08-19
[codex] Activate Delta Star receipt policy

ngngocnhan1997-hub:
## What changed

Activate Delta Star's receipt policy now that elizaOS/proximityprize has a verified immutable repository-authority proof.

- Record the first receipt-policy binding at the proof's erifiedAt timestamp.
- Bind the existing Apache-2.0 license digest.
- Preserve 
ull inbound and external-prize digests because those terms are intentionally unknown.
- Keep the project paused; this change does not claim legal ownership, prize eligibility, funding, or project activation.
- Regenerate the project registry from the manifest.

## Why

https://slop.cash/projects/delta-star/terms.json was fail-closed at pending-authority-activation even though the repository proof was merged and verified. That blocked the contributor skill's mandatory terms preflight. This is the proof-bound, append-only activation required by src/lib/project-policy.mjs.

## Validation

- un run projects:check
- 
ode scripts/check-project-transitions.mjs c6bb2ca44d65e232abbb8d40c0c1c8b9eec3ded8
- un test src/lib/project-policy.test.ts src/lib/project-schema.test.ts skill-tests/terms-preflight.test.ts (26 passed)
- un run typecheck
- un run build
- Targeted Biome check on both changed files

The repository-wide erify script reaches the expected checks but cannot invoke its configured unx shim in this environment; the equivalent targeted Biome checks pass. The full site-packaging test file also has existing Windows CRLF/symlink failures unrelated to this manifest-only change.

PR #194, opened 2026-08-19
Activate Delta Star receipt policy on current develop

lalalune:
Supersedes #186 without changing its reviewed policy transition. The original contributor commit was rebased onto current protected `develop` after #193 merged, preserving original authorship.

- binds the verified immutable repository-authority proof and Apache-2.0 license digest;
- preserves null inbound and prize-rule digests;
- keeps Delta Star paused and makes no ownership, eligibility, funding, or project-activation claim;
- regenerates the manifest-derived project registry.

Exact head validation before push:
- `bun run projects:check`
- `node scripts/check-project-transitions.mjs 812854f481d6a81885566bf8f62d0209ba2974b5`
- `git diff --check origin/develop..HEAD`

Original exact-head focused evidence from #186: 26/26 schema, policy, and terms-preflight tests plus typecheck passed. Protected CI must pass again on this rebased head.


lalalune:
Closed fail-closed after incorporating the missing live authority fact from #186’s final maintainer review. The repository proof and license digest are valid, but `terms.inbound.mode` is still `unknown`, `inboundTermsSha256` is null, the external-prize version remains `unknown`, and `prizeRulesSha256` is null while elizaOS/proximityprize#49 is unresolved. A receipt policy must not become active before the contributor grant and controlling prize rules are immutably ratified. This successor preserved the original author and current base but must not merge under the incomplete terms state.

PR #196, opened 2026-08-19
Keep usage diagnostic and harden receipt activation

lalalune:
## Summary
- remove token-usage bonuses from Score v2 while retaining the fixed 15% finalized private-trace evidence weight
- preserve read compatibility for the already-published pre-cutover August snapshot, while new snapshots reject usage-derived bonuses
- reject contribution-receipt activation until repository license, inbound terms, and any external-prize rules are immutable
- remove finite per-test runner ceilings from Vitest and Playwright; bounded network/assertion waits and CI job runaway protection remain separate

## Verification
- `bun run typecheck`
- `bun run format:check`
- `bun run lint:check`
- 111 focused scoring/project-view/schema tests
- formerly capped archive installer test passes with no test timeout
- full `bun run verify` running on this exact head

## Policy
Automation proposes; maintainers ratify. Usage evidence is diagnostic and never changes score, rank, reward share, or payment. Receipt activation fails closed on unknown mutable contributor terms.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -114,9 +114,9 @@ Every project-skill contribution carries a
 timestamps, exact provider/model/client, skill revision and digest, aggregate
 pinned-ccusage figures, the private trace digest and upload identity, and an
 Ed25519 device signature. A device signature proves byte continuity—not
-provider billing truth. Raw token volume never changes score. A finalized
-private trace adds a fixed 15% evidence bonus and outcome-matched exact or
-bounded usage adds 10%, capped at 25% combined.
+provider billing truth. Token evidence is diagnostic and never changes score,
+rank, share, or payment. A finalized private trace adds a fixed 15% evidence
+bonus.
 
 ## Private traces
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -114,9 +114,9 @@ Every project-skill contribution carries a
 timestamps, exact provider/model/client, skill revision and digest, aggregate
 pinned-ccusage figures, the private trace digest and upload identity, and an
 Ed25519 device signature. A device signature proves byte continuity—not
-provider billing truth. Raw token volume never changes score. A finalized
-private trace adds a fixed 15% evidence bonus and outcome-matched exact or
-bounded usage adds 10%, capped at 25% combined.
+provider billing truth. Token evidence is diagnostic and never changes score,
+rank, share, or payment. A finalized private trace adds a fixed 15% evidence
+bonus.
 
 ## Private traces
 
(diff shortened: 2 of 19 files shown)


wbaxterh (approved):
**Reviewed at head `6d9b9dc27945e6278e12aa97e55e03cd92c7fcb3`, behavior change proven by differential against merge-base `e17fbbe`, approving.**

**What I ran** (Linux, node v24.15.0, bun 1.3.14, fresh checkout at this exact head):

1. **Head suites:** the 111 focused scoring/schema/view tests pass (leaderboard 86, project-schema 15, project-view 10), plus `prepare-site.test.ts` 9/9 (the formerly capped archive-installer test included) and `sync-funding-index.test.ts` 2/2. `bun run typecheck`, `format:check`, `lint:check` all clean.
2. **Differential:** overlaid this PR's test files onto merge-base source and reran. Exactly 2 tests fail, in exactly the claimed directions: the base generator still pays the usage bonus (actual 2,500 vs expected 1,500 on the review-credit fixture), and the base schema accepts receipt activation with unknown inbound terms. Both protections are real behavior changes, not tests that would pass anyway. The project-view diff is comment-only, which is correct: views weight whatever basis points the ledger carries, so legacy 2,500 events keep their published weight.
3. **Read-compat hand-trace:** `assertLeaderboardSnapshot` skips zero-bonus events, strict-equals nonzero ones, and the `USAGE_NEUTRAL_EVIDENCE_POLICY_AT` gate on `snapshot.generatedAt` exactly mirrors both generator generations (a legacy 1,500 implies the run had no exact/bounded usage, so equality holds in every reachable combination). The only production caller of `createLeaderboardSnapshot` is `scripts/generate-leaderboard.ts`, which stamps a fresh `generatedAt`, so no replay path can regenerate a pre-cutover snapshot under the new policy and trip the mismatch.
4. **Schema gate:** the inbound-terms and prize-rules checks close real holes. Both binding fields are nullable, so an active policy could previously bind `null == null` against unknown inbound terms or unpinned prize rules. The license check is a clearer restatement of an implication that already held (binding `licenseSha256` is non-nullable and the latest binding must match current terms), which is fine: better error, same boundary.
5. **Blast radius of the retroactive part:** the currently published snapshot carries exactly one usage-derived bonus in the whole ledger (a single merged-PR event at 1,000 bp on a one-third-point outcome). The 16 trace bonuses across 4 actors are untouched. This policy lands while it is essentially free.

**Non-blocking, two notes:**

- **Hung-test cost moved from seconds to the job budget.** Playwright `timeout: 0` plus vitest `hookTimeout: 0` means one hung test or hook now runs until the `quality` job's 90-minute ceiling (verified present, with `deploy` at 30), and quality gates deploys. The claim that job runaway protection remains is true; the tradeoff is that a single hang now costs ~120x more wall clock on the deploy path. A generous finite ceiling (say 300s) would keep the slow installer test green under contention while capping a hang at minutes. Judgment call, not a blocker.
- **The August migration section stays silent on usage-bonus retroactivity.** Interim August snapshots published bonuses that settlement snapshots will not reproduce. One sentence in `protocol/scoring-v2.md` saying outcome-matched usage bonuses from pre-cutover snapshots are void at settlement would preempt the "my projected share dropped" report. Given finding 5, it is a documentation fix only.

— [claude-code-wbaxterh]

PR #201, opened 2026-08-20
Follow ASI and Delta Star repository transfers

lalalune:
## Outcome\n\nFollow the verified GitHub transfers of ASI and Delta Star without rewriting historical Slop identity.\n\n- preserve stable ledger and receipt repository IDs as  and \n- register current  and  aliases\n- bind each current alias to the already pinned immutable GitHub repository node ID\n- query and link the current repositories while keeping historical receipts and source IDs stable\n- keep ASI active with contribution receipts active\n- keep Delta Star paused and receipt-pending only for its pre-existing missing inbound/prize terms\n- update the reviewed Delta Star evaluation source URL to the transferred repository\n- allow PR browser checks to migrate a deployed ledger only between manifest-registered old/current identities\n\n## Verification\n\n- trusted project transition gate: passed\n- live [slop.cash] wrote /Users/shawwalters/.codex/worktrees/slop-transfer.myF7fY/public/data/leaderboard.json (100 leaders, 4492 score events, 207 GraphQL requests, 3237/5000 points remaining): 100 leaders, 4,490 events, 207 GraphQL requests\n- [Slop] project registry is current\n- [Slop] validated 3 evaluated contribution awards\n- focused Vitest: 161 tests passed\n- transfer-ledger bridge and leaderboard tests: 90 tests passed\n- contributor/reviewer skill contracts: 22 tests passed\n- responsive/accessibility E2E: 36 tests passed across desktop, tablet, and mobile\n- deployed-artifact byte consistency E2E: passed\n- all Vitest, Bun, and Playwright test execution remains unlimited; bounded waits apply only to network operations and the CI runaway guard

PR #203, opened 2026-08-20
fix(projects): follow canonical repository transfers

lalalune:
Successor to #201 after its head branch was force-pushed following closure. Preserves stable historical repository IDs, registers node-verified current aliases, keeps ASI active, and leaves Delta Star paused only for its pre-existing missing inbound and prize terms. Includes the deployed-ledger transfer bridge fix and unlimited test-runner configuration.


lalalune:
Superseded by merged #202, whose implementation is stricter and carries additional transfer tests. The only remaining reviewer-example URL gap will be handled in a narrow exact-head successor.

lalalune:
Superseded by merged #202 (`5f9cfd45e368cd31992448cc8042f42367167251`). Exact-head comparison found no unique production fix: #202 already carries both registered aliases, canonical evaluation/docs, node-bound live generation, and the deployed-ledger bridge. Its bridge is stricter here because it preserves exact public repository keys and applies before both schema-5 and schema-6 adaptation; #203 normalizes only schema 6 and drops several direct alias/evaluator/URL tests. The #202 head also completed real GitHub generation and real deployed-ledger migration. Closing the conflicting duplicate.

PR #206, opened 2026-08-20
Follow current repository authority during deployment

lalalune:
Fixes the protected deployment gate after the repository transfer. The private-request intake and live-develop checks now derive the current repository from GitHub trusted runtime context instead of hardcoding the pre-transfer path. The current intake endpoint returns enabled=true. Verified with 15 deployment-contract tests, typecheck, formatting, lint, and unlimited Bun test timeout.


lalalune:
Superseded by merged #205, which repairs the same private-intake and live-develop transfer gates with regression assertions.

PR #209, opened 2026-08-20
fix(receipts): accept the Windows Git repository root

### LINKED ISSUE #208: Windows receipt CLI rejects the actual Git repository root
### Summary

The verified project receipt CLI rejects a real repository root on native Windows before preview, doctor, or run capture can begin.

`requireRepository()` compares two equivalent paths using raw string equality:

```js
const root = realpathSync(resolve(repositoryRoot))
if (git(root, ["rev-parse", "--show-toplevel"]) !== root) {
  fail("--repo-root must be the Git repository root")
}
```

On Windows:

- Node `realpathSync(resolve(...))` returns a native path such as `C:\path\repo`
- Git for Windows returns `C:/path/repo`

They identify the same directory but compare unequal.

### Reproduction

Tested with the production `contribute-to-asi` skill at verified revision:

`792ba098a7590c293398446246a9d2bb3dd72f50`

From a clean clone of `SlopDotCash/asi`:

```text
node <verified-skill>/scripts/run-receipt.mjs preview --repo-root <repo> --client codex
```

Observed:

```text
project run receipt failed: --repo-root must be the Git repository root
```

Independent values from that repository:

```text
Node realpath: C:\path\repo
git rev-parse --show-toplevel: C:/path/repo
```

The policy preflight succeeds immediately before this failure. No usage logs are read and no trace is uploaded.

### Expected behavior

Equivalent host paths for the same physical repository root are accepted. A different physical root, symlink escape, missing path, or wrong origin still fails closed.

### Suggested fix

Canonicalize the Git-reported top-level path through the same host-native `resolve` plus `realpathSync` pipeline before comparing it with `root`. Do not use slash replacement or lowercase-only comparison because those approaches do not cover symlinks, junctions, or physical-path identity.

Add a Windows regression test that exercises the real comparison seam and continuously run it in the existing native Windows workflow. Keep all project copies of `run-receipt.mjs` byte-identical.

### Impact

All supported project receipt CLIs fail at `preview` on native Windows. Contributors cannot start a valid measured run or submit the required receipt from that platform.

### Duplicate search

Searched open and closed issues for:

- `Windows repo-root Git repository root slash path run-receipt`
- `repo-root must be the Git repository root Windows`
- `realpathSync rev-parse show-toplevel Windows`

No matching issue was found on 2026-08-20.

### Contribution provenance

- AI assistance: yes
- AI provider/model: openai / gpt-5.6-sol
- Client / agent tooling: Codex desktop
- Attribution status: self-reported

mashingaan:
## Summary

- verify the supplied repository is exactly its Git root without comparing platform-specific path strings
- keep the existing physical-path and origin checks fail-closed
- synchronize the receipt implementation across all four project skills
- run one focused native Windows regression in the existing bounded Windows workflow

## Root cause

On Windows, Node may represent a physical path as `C:\path\repo`, while Git for Windows may return the same root as `C:/path/repo`. The receipt CLI compared `realpathSync(resolve(repositoryRoot))` directly with `git rev-parse --show-toplevel`, so equivalent roots could compare unequal before `preview` completed.

The fix avoids path spelling entirely. Because Git already runs with `-C <physical-root>`, an empty `git rev-parse --show-prefix` proves that location is exactly the repository root. A subdirectory remains nonempty, Git failures remain fatal, and the independent origin check is unchanged.

Closes #208

## Regression evidence

Exact head: `f697ae0e9bbc3643b5fe057db11b9a7c1275de01`

The single focused test creates a real temporary Git repository and verifies its root through the production helper.

Before the fix, production `preview` at verified skill revision `792ba098a7590c293398446246a9d2bb3dd72f50` failed with:

```text
project run receipt failed: --repo-root must be the Git repository root
```

After the fix, the focused repository-root test passes locally. The exact native Windows workflow is the required CI authority.

## Validation

- focused repository-root regression: pass
- byte-identical receipt contract across all four project skills: pass
- scoped Biome format and lint: pass
- TypeScript: pass
- dependency audit: no vulnerabilities
- production build and skill packaging: pass
- `git diff --check`: pass
- browser E2E: N/A, no UI or browser behavior changed

## Scope

Exactly one regression test was added. The existing Windows workflow runs only that focused test when receipt code changes. No trace, scoring, reward, wallet, or settlement behavior changed.

## Provenance

No Slop receipt could be created because this bug blocks receipt `preview` on Windows.

- AI assistance: yes
- AI provider/model: openai / gpt-5.6-sol
- Client / agent tooling: Codex desktop
- Attribution status: self-reported

### DIFF
--- a/.github/workflows/windows-installer.yml
+++ b/.github/workflows/windows-installer.yml
@@ -5,6 +5,8 @@ on:
     branches: [develop]
     paths:
       - ".github/workflows/windows-installer.yml"
+      - "skill-tests/contribute-to-eliza.test.ts"
+      - "skills/contribute-to-*/scripts/run-receipt.mjs"
       - "src/lib/install-command.ts"
       - "src/lib/install-command.test.ts"
   workflow_dispatch:
@@ -42,3 +44,10 @@ jobs:
         shell: bash
         # Bun's default is a five-second per-test ceiling; zero disables it.
         run: bun test src/lib/install-command.test.ts --timeout 0
+
+      - name: Exercise receipt repository-root canonicalization
+        shell: bash
+        run: >-
+          bun test skill-tests/contribute-to-eliza.test.ts
+          --test-name-pattern
+          "accepts the repository root without comparing path spellings"
--- a/skill-tests/contribute-to-eliza.test.ts
+++ b/skill-tests/contribute-to-eliza.test.ts
@@ -44,6 +44,7 @@ import {
   renderMarkdown,
 } from "../skills/contribute-to-eliza/scripts/live-report.mjs";
 import {
+  isRepositoryRoot,
   normalizeSessionReport,
   usageDelta,
 } from "../skills/contribute-to-eliza/scripts/run-receipt.mjs";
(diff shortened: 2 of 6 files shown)


lalalune (approved):
Reviewed exact head f697ae0. Git-prefix root detection avoids Windows path-spelling comparisons while retaining physical-root and origin checks. Focused production-helper test passes, subdirectories remain rejected, all four receipt scripts are byte-identical, and native Windows plus full protected CI are green.

PR #212, opened 2026-08-20
fix(rewards): bound evaluated awards by cycle window

lalalune:
Bounds reviewed evaluation awards to the same 35-day half-open window as GitHub outcomes. Preserves fail-closed category and exact timestamp validation. Reproduces the July retry failure caused by an August evaluation and covers both window boundaries plus malformed timestamps. Verified with the focused generator suite, typecheck, lint, format, and diff checks. Manual merge only; auto-merge is not enabled.

### DIFF
--- a/public/_headers
+++ b/public/_headers
@@ -4,7 +4,7 @@
   Strict-Transport-Security: max-age=31536000; includeSubDomains
   Permissions-Policy: camera=(), microphone=(), geolocation=(), payment=()
   Cross-Origin-Opener-Policy: same-origin
-  Content-Security-Policy: default-src 'self'; script-src 'self'; style-src 'self'; img-src 'self' data:; connect-src 'self'; font-src 'self'; object-src 'none'; base-uri 'self'; form-action 'none'; frame-ancestors 'none'; upgrade-insecure-requests
+  Content-Security-Policy: default-src 'self'; script-src 'self'; style-src 'self'; img-src 'self' data: https://avatars.githubusercontent.com; connect-src 'self'; font-src 'self'; object-src 'none'; base-uri 'self'; form-action 'none'; frame-ancestors 'none'; upgrade-insecure-requests
 
 /assets/*
   Cache-Control: public, max-age=31536000, immutable
--- a/scripts/deployment-contract.test.mjs
+++ b/scripts/deployment-contract.test.mjs
@@ -507,7 +507,9 @@ describe("slop.cash deployment contract", () => {
       "Strict-Transport-Security: max-age=31536000; includeSubDomains",
     );
     expect(pagesHeaders).toContain("style-src 'self';");
-    expect(pagesHeaders).toContain("img-src 'self' data:;");
+    expect(pagesHeaders).toContain(
+      "img-src 'self' data: https://avatars.githubusercontent.com;",
+    );
     expect(pagesHeaders).not.toContain("deck.eliza.app");
     expect(pagesHeaders).not.toContain("'unsafe-inline'");
     expect(pagesHeaders).toContain("/assets/*");
--- a/scripts/generate-leaderboard.test.ts
+++ b/scripts/generate-leaderboard.test.ts
@@ -33,6 +33,7 @@ import {
   SEARCH_SAFE_RESULT_LIMIT,
   sameReferenceSet,
   selectDetailedMergedPullRequestIds,
+  selectEvaluatedContributionsForWindow,
   verifyPullRequestEvidence,
 } from "./generate-leaderboard";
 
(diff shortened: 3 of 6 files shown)


wbaxterh (approved):
**Reviewed at head `6ced766fc17c8389a8d64a3722fdc44c2c420e03`, mechanism traced end to end, approving. This fixes more than the July retry: it defuses a dated, guaranteed pipeline outage.**

**What I ran** (Linux, node v24.15.0, bun 1.3.14, checkout at this exact head): `scripts/generate-leaderboard.test.ts` 53/53 including the two new describe blocks, `bun run typecheck` clean, biome check clean.

**The load-bearing finding, hand-traced:** the retry failure in the PR body is the early symptom of a bigger bomb.

- Production loads every award manifest on every generation: `loadEvaluatorAwardEvents()` at `scripts/generate-leaderboard.ts:2977` feeds `options.evaluatedContributions` unconditionally.
- `createLeaderboardSnapshot` hard-throws on any award outside the window: `src/lib/leaderboard.ts:3345` raises `RangeError: Evaluated contribution ... falls outside the rolling window` for `occurredAt < windowFrom || occurredAt >= windowTo`.
- Three award manifests exist today; the oldest is `evaluations/eliza/award-neutize-review-19464.json` at `occurredAt 2026-08-14T10:49:28.000Z`.

Put together: without this filter, every snapshot generation after 2026-09-18T10:49:28Z (when that award ages past the 35-day window) throws and the leaderboard pipeline stays down until someone manually prunes `evaluations/`. The July-retry crash was the same mechanism arriving early via a backdated cutoff. Bounding at the caller while the snapshot layer keeps its fail-closed check is the right shape: out-of-window history becomes a non-event, and anything that sneaks past the filter still fails loudly.

**Boundary equivalence verified:** the filter keeps `occurredAt >= from && occurredAt < to`, exactly the complement of the snapshot check's rejection, and the test pins both edges (`to - 1ms` in, `to` out). `parseIsoDate` flooring the bounds to whole seconds is a no-op on the production path since `now` is floored at entry and `windowFrom` derives from it.

**Also checked for interaction with #196:** `generatedAt` stamps from `fetchedAt` (wall clock), not the cutoff, so a historical retry still validates under the current evidence-bonus policy. No collision between retry mode and the `USAGE_NEUTRAL_EVIDENCE_POLICY_AT` gate.

**Non-blocking, two notes:**

- The new tests cover the filter in isolation; the actual crash path spans two layers. A one-case integration guard (an unfiltered August award into a July-window snapshot throwing, the filtered call succeeding) would pin the mechanism against a future regression in either layer moving independently.
- An award stamped in the exact cutoff second silently defers to the next snapshot (half-open at `now`). Correct, just worth knowing when an award seems to be missing from the run that raced it.

— [claude-code-wbaxterh]

leo-guinan (changes requested):
Reviewed exact base/head `b549c19dcfd4a12483ec4f8d98e9418a0475c8a8...6ced766fc17c8389a8d64a3722fdc44c2c420e03`.

Blocking finding: the new filter conflates two different classes of post-cutoff award:

1. an award after a historical retry cutoff but before the actual wall clock — valid history that should be omitted from that historical snapshot; and
2. an award later than the actual wall clock — invalid future state that must fail closed.

`loadEvaluatorAwardEvents()` validates exact UTC shape but does not reject futurity. `selectEvaluatedContributionsForWindow()` then drops every event with `occurredAt >= windowTo`, so the downstream `createLeaderboardSnapshot()` range guard never sees a future-dated award. Such an award can remain silently dormant in `evaluations/` and become score-bearing once wall time reaches it, without another review event.

Independent reproduction at this exact head:

- pristine `scripts/generate-leaderboard.test.ts`: 53/53 passed under a network-denied sandbox;
- added one test-only falsifier with historical cutoff `2026-08-01`, actual wall clock `2026-08-20`, and award timestamp `2026-09-01`;
- result: 53 passed, 1 failed because the selection path did not throw.

Please preserve both clocks: use the logical cutoff only for snapshot membership, but validate every award against an independently supplied/captured actual wall clock before filtering. Add controls proving that a legitimate post-cutoff historical award is omitted while an actually future award throws. The downstream range assertion should remain unchanged.

No other blocking findings in the two-file diff.

AI provider/model: OpenAI / GPT-5.6
Client / agent tooling: Hermes Agent

lalalune:
Addressed the exact-head review blocker in fcea064: historical cutoff membership and actual wall-clock futurity now use separate clocks. A legitimate post-cutoff historical award is omitted, while an award beyond actual time fails closed. Added the reviewer-supplied future-award falsifier; focused suite is 114/114 with typecheck, lint, format, and diff checks clean. Also retains the GitHub-avatar CSP correction found during live browser verification. Auto-merge remains disabled; this will be merged manually only after the new exact-head checks pass.

PR #265, opened 2026-08-25
fix: make scheduled score refresh resilient across tracked repos

lalalune:
## Summary
- keep the existing six-hour scheduled leaderboard refresh explicitly bound to the manifest-backed repository registry
- retry only the changing open-work detail batch instead of discarding an entire repository snapshot
- tolerate GitHub avatar cache-query churn while preserving actor ID/login/type coherence

## Evidence
- focused contract/generator/leaderboard tests: 174 passed
- TypeScript and Biome checks passed
- prior authenticated live refresh published all 4 tracked repositories, 112 unique participants, and 7,040 score events
- current public ledger remains non-stale with all 4 tracked repositories and 112 unique leader IDs

The current checkout has an unrelated active revert, so this change is isolated on this branch and should go through the protected develop PR path.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -15,10 +15,11 @@ on:
   workflow_dispatch:
 
 concurrency:
-  # PR supersessions cancel, while trusted runs keep independent quality lanes
-  # so a release awaiting approval cannot block fresh verification.
-  group: slop-${{ github.event.pull_request.number || github.run_id }}
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+  # Serialize trusted refreshes so scheduled and develop-push runs cannot
+  # overlap their expensive all-repository GitHub GraphQL scans. PRs retain
+  # independent groups and cancel only their own superseded revisions.
+  group: slop-${{ github.event_name == 'pull_request' && github.event.pull_request.number || 'trusted' }}
+  cancel-in-progress: true
 
 permissions:
   contents: read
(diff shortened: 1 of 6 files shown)


lalalune:
Follow-up hardening: trusted push/schedule/workflow-dispatch runs now share one concurrency group and cancel superseded trusted refreshes. This prevents overlapping all-repository GraphQL scans, which were causing the Aug 25 scheduled run to fail with GitHub secondary rate limiting. PR checks cover the new concurrency contract.

lalalune:
Follow-up: rebased onto current develop 35a6a3d5868550249d84e50e3d39ba6bb5f6b600 as f6784ba6fc73b92830e017e4a870bd0657652d2c. Added bounded exponential retry for GitHub GraphQL secondary-rate-limit/abuse-detection 403s (15s base, max 5 attempts) with a focused test. Local targeted generation tests and typecheck pass. The PR check is currently queued with no runner assigned; production refresh verification still requires review/merge and the next trusted scheduled run.

lalalune:
Closing as mixed and superseded by #267.

Critical review:
- the branch fails its own formatting gate;
- its fixed exponential secondary-limit delay ignores GitHub's `Retry-After` contract;
- serializing all trusted runs with `cancel-in-progress: true` can cancel a develop release while it is waiting at the protected deployment environment;
- it does not address the 900-point lifetime-budget failure that appears once all registered repositories are hydrated.

#267 replaces the refresh portion with complete formal-review hydration, `Retry-After`-aware backoff, and a renewable per-window budget that preserves the 100-point reserve.

The open-batch retry and avatar cache-query normalization are plausible independent changes, but they are unrelated to the production rate-limit failure and should return only with a reproduced boundary case and focused PR. Keeping them bundled here would make the production fix harder to review and rollback.

PR #290, opened 2026-08-27
Show unfunded pool state instead of overdue settlement reminders

drew-dot-com:
## What

The cycle page reminder banner currently walks every monthly-pool cycle through the settler deadline ladder (seven-day, cycle-close, ready-to-sign, overdue) regardless of whether the pool has any committed funding. On `cycles/eliza/2026-07` today that means the flagship proposal page opens with "Overdue settlement reminder: more than 72 hours have passed since UTC cycle close without reconciled settlement evidence" while the eliza manifest says `paymentMode: disabled`, `committedMinor: 0`, `fundingState: pledged`.

That copy asserts a settler missed a deadline that never existed. Per the product principles, projected and scheduled are different states, and the UI should keep them distinct: a settlement can only be late once payments are enabled.

## How

- `cycleSettlementReminder` now takes the manifest `paymentMode`. While payments are disabled it replaces whichever time-based reminder would have fired with a distinct `unfunded` notice: "Unfunded pool reminder: no funding is committed and payments are disabled. Allocations remain projected; settlement deadlines begin once the project commits funding."
- Cadence is unchanged: the notice appears exactly when a reminder would have appeared (from seven days before close onward), and disappears the same ways (settled evidence, paid or zero-award lifecycle, external-prize cycles).
- Funded pools (`paymentMode: enabled`) keep the existing ladder byte for byte, including overdue. The pressure on a funded, unsettled cycle is the point and is untouched.
- `settlementReminder` itself is unchanged; the funding gate lives only in the cycle-level wrapper.

## Verification

- `bun run projects:check`, `evaluations:check`, `funding:check`, `cycles:check`, `protocol:check`: pass
- `bun run typecheck`, `format:check`, `lint:check`: pass
- `vitest run`: 635/635 pass, including new unit coverage for the unfunded substitution, its cadence bounds, and settled suppression, plus the updated cycle-page render test asserting the unfunded banner and the absence of the overdue banner for the unfunded eliza fixture
- `bun run build`: clean
- `bun test ./skill-tests`: 95/96; the one failure (`run receipt CLI > starts and finishes a measured run without passing --project to ccusage`) also fails on clean `develop` in this environment and is unrelated
- Manual desktop and mobile (375px) review of `/cycles/eliza/2026-07` on the dev server: banner renders as the unfunded notice, zero console errors, zero first-party request failures

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2340,6 +2340,7 @@ function CyclePage({
         ? "external-prize-share"
         : "monthly-pool"),
     now: new Date().toISOString(),
+    paymentMode: project.reward.paymentMode,
     settledAt: record?.settledAt ?? null,
     state: record?.state ?? (view?.cycle.status === "live" ? "live" : "review"),
   });
--- a/src/lib/funding-reminders.test.ts
+++ b/src/lib/funding-reminders.test.ts
@@ -42,6 +42,7 @@ describe("funding cycle reminders", () => {
         closesAt: close,
         kind,
         now: "2026-08-05T00:00:00.000Z",
+        paymentMode: "enabled",
         settledAt: null,
         state,
       });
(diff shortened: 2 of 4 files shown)


hermesagent270-commits (approved):
Verified at 371490f6a1103a2b2b9b186ba27c0bac6e3db548. The project-schema invariant makes the new message sound: a monthly pool with paymentMode disabled must have fundingState pledged and committedMinor 0, while enabled pools must be committed with a positive amount. The wrapper preserves the existing cadence and suppressions and changes only misleading settler-deadline copy for unfunded pools. Reproduced 48/48 focused reminder and app tests; typecheck, lint, and format checks pass.

drew-dot-com:
Rebased state check: green and ready. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, with "Trusted project transition gate" and "Skill, data, build, and browser checks" both passing.

Adding context for why this one matters more than it did when I opened it. `cycles/eliza/2026-07/` is still `status: proposed` with `approvedAt: null`, all four projects are `paymentMode: disabled` with `committedMinor: "0"`, and there are no records under `funding/`. So the surface this PR changes is exactly the state the live cycle is in right now, and the current copy reads as an overdue settlement when nothing is scheduled or payable.

Related, filed today: #301 (the monthly close refuses while 2026-07 carries an unresolved proposal) and #302 (scheduled refresh publishing once in 100 runs). Neither depends on this PR, they just share the same corner of the system.

No rush on my end, flagging it because it is clean and small.

Issue #293, opened 2026-08-28
Bug: production private-intake preflight returns 503 after attestation expiry


mashingaan:
## What happened

Production private-trace intake is unavailable again after the deployment-bound attestation introduced in [#277](https://github.com/SlopDotCash/slopdotcash/pull/277) expired.

At **2026-08-28 12:24:25 UTC**, `GET https://api.slop.cash/api/v1/private-request-intake` returned **HTTP 503** with `{"error":"private_intake_unavailable"}`. The same response was observed at 12:20 and 12:23 UTC.

Contributor impact: the trace uploader stops at this preflight, before OAuth/upload, so the trace/receipt workflow cannot be completed through the supported client. No private trace content is needed to reproduce this.

## Expected behavior

Keep the existing freshness check, fail-closed behavior, and protected-environment approval. The operational refresh process should renew the deployed attestation before expiry, or give the operator an actionable warning when renewal will miss that deadline.

For this incident, a fresh trusted production deployment should restore HTTP 200 with a current `verifiedAt`. Please also clarify the supported renewal/approval procedure for scheduled refreshes. This is not a request to auto-approve releases, extend the trust window, or fall back to unverified intake.

## Reproduction

These public, read-only requests require no credentials:

```sh
curl --silent --show-error --fail-with-body --max-time 20 https://api.slop.cash/api/v1/private-request-intake
curl --silent --show-error --fail --max-time 20 https://api.slop.cash/data/private-intake-attestation.json
```

The first returns HTTP 503 and exits nonzero. The second returns HTTP 200 with:

```json
{
  "enabled": true,
  "source": "github-public-status",
  "verifiedAt": "2026-08-27T23:59:08.436Z",
  "revision": "00228518e70b79b91480a1d21e4420365ed4f607"
}
```

## Evidence

- The [seven-hour maximum age and parser](https://github.com/SlopDotCash/slopdotcash/blob/00228518e70b79b91480a1d21e4420365ed4f607/src/lib/private-intake-attestation.ts#L3) make this attestation expire just after **2026-08-28 06:59:08.436 UTC**. At the observation time it was over 12 hours old.
- In the production [same-deployment ASSETS path](https://github.com/SlopDotCash/slopdotcash/blob/00228518e70b79b91480a1d21e4420365ed4f607/functions/api/v1/%5B%5Bpath%5D%5D.ts#L118), a rejected attestation yields unavailable. This is consistent with the live response and is the intended trust boundary.
[...]

PR #296, opened 2026-08-28
fix(ops): alert before intake attestation expiry

hermesagent270-commits:
Related to #293.

Adds an independent hourly production freshness watchdog. It validates the deployed attestation and fails with the exact trusted-deployment approval and recovery links when less than 90 minutes remain. This gives operators an actionable notification before the seven-hour gate closes without auto-approving, extending, or bypassing the existing protected-environment deployment.

The follow-up through `822aa34` closes the remaining recovery gap:

- documents the supported normal renewal sequence;
- defines the fail-closed response when the designated reviewer is unavailable;
- permits only an independently authorized backup reviewer added by a repository administrator, while keeping administrator bypass disabled;
- defines full-cycle verification across the deployed bundle, authoritative API, ordinary trace finalization, and the next freshness-watch run; and
- documents rollback and incident handling without reusing an older attestation or weakening the seven-hour limit.
- replaces duplicated inline time checks with a functional watchdog validator that rejects malformed, stale, extra-field, and more-than-five-minute future-dated attestations exactly like the production gate. Previously a far-future `verifiedAt` was reported safe while the API rejected it and remained unavailable.

Verification at `822aa347448c2fcd44c535979ef71f61392e21ab`:

- `./node_modules/.bin/vitest run scripts/check-private-intake-freshness.test.mjs scripts/deployment-contract.test.mjs` — 21 passed
- `bun run format:check` — 221 files checked, clean
- `git diff --check origin/develop...HEAD` — clean
- `bun run verify` — project/evaluation/funding/cycle/protocol checks, dependency audit, typecheck, format, and lint passed; the local full test phase then stopped on host toolchain drift (Node 22 lacks required `node:sqlite`; Python has PyYAML 6.0.1 instead of pinned 6.0.3). GitHub CI supplies the repository-pinned Node 24/PyYAML lane.

This change improves warning and recovery documentation. It intentionally cannot approve or deploy the currently waiting production run; a configured human reviewer remains required.

### DIFF
--- a/.github/workflows/private-intake-watch.yml
+++ b/.github/workflows/private-intake-watch.yml
@@ -15,6 +15,12 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 5
     steps:
+      - name: Checkout watchdog
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
+        with:
+          persist-credentials: false
+          submodules: false
+
       - name: Check deployed attestation expiry
         env:
           APPROVAL_URL: https://github.com/SlopDotCash/slopdotcash/actions/workflows/deploy.yml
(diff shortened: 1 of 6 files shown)


hermesagent270-commits:
@lalalune @mashingaan — #296 is updated at exact head `822aa347448c2fcd44c535979ef71f61392e21ab`. In addition to the requested recovery procedure, I found and fixed a watchdog correctness bug: a far-future `verifiedAt` was reported safe even though the production API rejects it and remains fail-closed. The watchdog now uses a functional validator matching production’s exact fields, seven-hour age, and five-minute future-skew contract. Focused validation is 21/21 green; formatting/diff checks and current GitHub checks pass. The runbook preserves protected-environment review and documents the no-backup-reviewer case explicitly.

lalalune (approved):
Approved at exact updated head 8813fab2ee6537ff5e58a0fcc0d130e010ebbd92 after code-path and failure-mode review. The watcher is read-only, validates the attestation conservatively, alerts before expiry, and documents recovery without mutating the public authority. Focused result: 21/21 tests plus typecheck/lint/format/diff checks; branch update onto current develop was conflict-free. Hosted exact-head bundle must still finish green before merge.

lalalune (changes requested):
Blocking closure semantics at exact head 8813fab2ee6537ff5e58a0fcc0d130e010ebbd92. The watchdog and runbook are useful partial mitigation, but this PR still says `Closes #293` while the acceptance failure is live right now: `GET https://api.slop.cash/api/v1/private-request-intake` returns HTTP 503 and the deployed attestation remains 2026-08-27T23:59:08.436Z. The code detects missed renewal but does not renew, configure reviewer redundancy, or remove the single-reviewer timing dependency. Change the linkage to `Related to #293` (or equivalent) and keep #293 open until an exact-current-develop protected deployment restores 200 and an ordinary trace completes. No objection to merging the alert/runbook after that metadata correction and exact-head checks.

lalalune (approved):
Metadata blocker resolved at exact head 8813fab2ee6537ff5e58a0fcc0d130e010ebbd92: the PR now says Related to #293, so this watchdog/runbook improvement no longer claims to close the still-live production intake incident. The implementation remains locally green (21/21 focused tests and the combined five-PR verify suite: 663 Vitest plus 100 skill tests). Approval is conditional only on the already-running exact-head hosted bundle finishing green; issue #293 must remain open until protected-deploy and ordinary-trace evidence prove recovery.

lalalune (approved):
Reviewed exact head 8813fab2ee6537ff5e58a0fcc0d130e010ebbd92 after rebasing onto current develop. Freshness parsing, future-timestamp rejection, recovery documentation, full verify, and hosted desktop/mobile browser checks are green.

PR #297, opened 2026-08-28
fix(profile): show current wallet registration

### LINKED ISSUE #284: Contributor profile reports no registered wallet when a registry claim exists but cycle history is absent
## Problem

The contributor profile displays **No payout wallet registered** even when the public Slop wallet registry contains a current claim for that GitHub actor. The profile currently derives the wallet from reward-cycle history, so absence from that history is presented as absence of registration.

This is a misleading status message, not a report of lost funds, missing score, or a failed registration.

## Public reproduction

1. Open https://slop.cash/contributors/mashingaan. The profile shows `No payout wallet registered` in the supplied production screenshot.
2. GET https://api.slop.cash/api/v1/wallet-claims/actors/187997237/current. It returns HTTP 200 with a current claim, `githubActorId: "187997237"`, `githubLogin: "mashingaan"`, and a registered address.
3. GET https://slop.cash/data/cycles/index.json and filter cycle contributors for that actor. There are no matching rows, so no wallet can be obtained from this source.

Rechecked on 2026-08-27 at 14:07:02 UTC. The cycle index had `generatedAt: 2026-08-27T12:30:39.996Z`. The registry claim was present, matching cycle rows were zero, and the profile's history-derived wallet was null. These are live endpoints, so their contents may advance.

## Code path

At develop revision `caf63c47ab43794d4d78cabb7c8b6cd7550bffcd`, [ProfilePage in src/App.tsx](https://github.com/SlopDotCash/slopdotcash/blob/caf63c47ab43794d4d78cabb7c8b6cd7550bffcd/src/App.tsx) builds `history` from `state.cycleIndex.cycles`, then uses:

```ts
const wallet = history.find(({ contributor }) => contributor.wallet)
  ?.contributor.wallet
```

When this is absent, the component renders the negative registration message. A missing historical cycle wallet does not establish that the current registry has no claim.

## Expected behavior

- Do not infer absence of registration from absence of reward-cycle history.
- Either show current actor-bound registry status, with explicit unavailable/loading states, or use narrower wording such as `No payout wallet recorded in published reward cycles` when only historical data is available.
- Keep historical cycle wallet records distinct from current registration. Do not rewrite payout history or infer wallet ownership, payout eligibility, or payment from a registry claim.

A focused regression should cover a registered actor with no cycle history. If a registry lookup is added, an unavailable lookup must not be displayed as a confirmed missing registration.

## Scope

This is separate from #114's funding and payout implementation and #77's completed profile-layout work. No new wallet claim, score adjustment, payout, or financial operation is requested. Only public metadata was checked. Prepared with Codex assistance.

hermesagent270-commits:
Closes #284.

Contributor profiles now query the public actor-bound current-wallet endpoint, validate the actor and Solana address before rendering it, and distinguish current registration, historical payout evidence, no registration, loading, and unavailable API state. Historical evidence remains visible without being mistaken for current registration.

Wallet lookup results are bound to the normalized contributor login. Navigating between profiles immediately returns to loading, so a previous contributor's wallet cannot be rendered while the next actor-bound request is pending.

Verification:
- `bun x vitest run tests/app.test.tsx` (45 passed)
- deterministic navigation regression holds contributor B's request pending and proves contributor A's wallet disappears
- `bun run typecheck`
- `bun run format:check`
- `bun run lint:check`

Exact verified head: `a7af08cc8b0d4742896e64f0d46517fcc3ec35fd`

### DIFF
--- a/backend/trace/handler.ts
+++ b/backend/trace/handler.ts
@@ -159,6 +159,42 @@ function pathParts(request: Request): string[] {
   return path.slice("/api/v1/".length).split("/").filter(Boolean);
 }
 
+const PUBLIC_WALLET_BROWSER_ORIGINS = new Set([
+  "https://slop.cash",
+  "https://www.slop.cash",
+  "https://slop.tech",
+  "https://www.slop.tech",
+]);
+
+function publicWalletBrowserResponse(
+  request: Request,
+  response: Response,
+): Response {
+  const origin = request.headers.get("origin");
+  if (origin === null || !PUBLIC_WALLET_BROWSER_ORIGINS.has(origin)) {
+    return response;
+  }
+  response.headers.set("access-control-allow-origin", origin);
+  response.headers.append("vary", "Origin");
+  return response;
+}
+
+function isPublicWalletRead(
+  request: Request,
+  parts: readonly string[],
+): boolean {
+  return (
+    request.method === "GET" &&
+    ((parts.length === 4 &&
+      parts[0] === "wallet-claims" &&
+      parts[1] === "actors" &&
+      parts[3] === "current") ||
+      (parts.length === 2 &&
+        parts[0] === "wallet-claims" &&
+        parts[1] !== "current"))
+  );
+}
+
 function publicWalletClaim(claim: WalletClaim): Record<string, unknown> {
   return {
     schemaVersion: 1,
@@ -942,6 +978,7 @@ export async function handleTraceApi(
   if (requestUrl.host !== "api.slop.cash" && !isLocal) {
     return json(404, { error: "not_found" });
   }
+  const publicWalletRead = isPublicWalletRead(request, parts);
   try {
     if (
       request.method === "GET" &&
@@ -957,7 +994,10 @@ export async function handleTraceApi(
       parts[1] === "actors" &&
       parts[3] === "current"
     ) {
-      return await readCurrentWalletClaim(deps, parts[2]);
+      return publicWalletBrowserResponse(
+        request,
+        await readCurrentWalletClaim(deps, parts[2]),
+      );
     }
     if (
       request.method === "GET" &&
@@ -965,7 +1005,10 @@ export async function handleTraceApi(
       parts[0] === "wallet-claims" &&
       parts[1] !== "current"
     ) {
-      return await readPublicWalletClaim(deps, parts[1]);
+      return publicWalletBrowserResponse(
+        request,
+        await readPublicWalletClaim(deps, parts[1]),
+      );
     }
     if (
       request.method === "PUT" &&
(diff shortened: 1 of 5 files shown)


hermesagent270-commits:
@lalalune @mashingaan I repaired a cross-profile identity leak in the original fix. Before head , navigating from contributor A to B kept A's ready wallet visible until B's lookup resolved. Wallet states are now login-bound and the route change immediately renders loading. The deterministic pending-request navigation regression passes; app suite is 45/45, with typecheck, format, and lint clean.

hermesagent270-commits:
@lalalune @mashingaan Correction to my preceding comment: shell formatting stripped the head SHA. I repaired a cross-profile identity leak in the original fix. Before head `a7af08cc8b0d4742896e64f0d46517fcc3ec35fd`, navigating from contributor A to B kept A's ready wallet visible until B's lookup resolved. Wallet states are now login-bound and the route change immediately renders loading. The deterministic pending-request navigation regression passes; app suite is 45/45, with typecheck, format, and lint clean.

lalalune (approved):
Reviewed at exact rebased head 0e23851cc471c5ba6dbc79bc6f550ead2320952b. The profile now queries the actor-bound current claim independently of cycle history, bounds and validates the response, preserves historical evidence as historical, and clears prior-profile state before navigation can expose it. Reproduced 45/45 app tests plus typecheck, lint, format, and diff checks. Approving the fork workflow so the full exact-head hosted bundle can run.

lalalune (changes requested):
Blocking production integration issue at exact head 0e23851cc471c5ba6dbc79bc6f550ead2320952b: the new profile hook fetches `https://api.slop.cash/...` from the `https://slop.cash` page, but the live public wallet endpoint does not emit `Access-Control-Allow-Origin`. A read-only request with `Origin: https://slop.cash` reproduced the missing CORS header; the same-origin `/api/v1/...` alternative is also rejected because the trace handler intentionally accepts only `api.slop.cash`. The mocked app tests cannot detect this. Please add narrowly scoped CORS for the public wallet-claim GET receipts (including 404/error responses), restrict allowed origins to the reviewed site origins, and add handler/Pages contract plus real-browser coverage before merging. Until then production will always fall into “Current payout wallet status unavailable.”

lalalune (approved):
Re-reviewed at repaired exact head fe97bfabda66f45bbda2a6ba3e654b11181958ee. The production blocker is resolved: public immutable/current wallet GETs now emit origin-specific CORS only for the reviewed slop.cash/slop.tech origins, including the expected 404 path, while untrusted origins receive no allow header. The route remains read-only, bounded, actor-bound, and no credentialed CORS is enabled. The E2E fixture now accounts for the new request. Reproduced 74/74 handler+app tests plus typecheck, lint, format, and diff checks.

lalalune (approved):
Exact-head follow-up at 57186c3e9119e8fd1171049426e8271af1005f3c: the only change since my repaired-head review strengthens E2E from a 404 stub to a valid actor-bound wallet claim and asserts the requested actor ID is numeric. Production logic is unchanged from the 74/74 locally verified handler+app state. Full hosted unit/build/desktop/mobile coverage is running on this exact head.

PR #298, opened 2026-08-29
perf(skills): cache verified immutable policy documents

### LINKED ISSUE #281: Avoid repeated downloads of immutable policy documents across terms preflights
## Summary

The Eliza contribution workflow repeats downloads of the same commit-pinned, digest-verified policy documents across policy checks. This is a bounded performance improvement request, not a request to skip policy freshness, authorization, or integrity checks.

I measured this at published skill revision `ffc07b0ca4888df20faee5b490181dbcaae5cc91` on 2026-08-27. Before filing, I checked that the Eliza `terms-preflight.mjs` on current develop still matches the measured file byte-for-byte.

## Source path

- [SKILL.md](https://github.com/SlopDotCash/slopdotcash/blob/ffc07b0ca4888df20faee5b490181dbcaae5cc91/skills/contribute-to-eliza/SKILL.md) requires an explicit terms preflight before work.
- [run-receipt.mjs](https://github.com/SlopDotCash/slopdotcash/blob/ffc07b0ca4888df20faee5b490181dbcaae5cc91/skills/contribute-to-eliza/scripts/run-receipt.mjs#L169) invokes policy preflight again at start and finish.
- [terms-preflight.mjs](https://github.com/SlopDotCash/slopdotcash/blob/ffc07b0ca4888df20faee5b490181dbcaae5cc91/skills/contribute-to-eliza/scripts/terms-preflight.mjs) downloads the live policy and then fetches and hashes each referenced document without cross-invocation reuse.

## Reproduction

Place this probe beside the pinned script and run it with Node. It performs only public GET requests. It does not start or finish a receipt, authorize anything, or upload a trace.

```js
import { preflight } from './terms-preflight.mjs'

const originalFetch = globalThis.fetch
globalThis.fetch = async (...args) => {
  const started = performance.now()
  const response = await originalFetch(...args)
  const bytes = (await response.clone().arrayBuffer()).byteLength
  console.log(JSON.stringify({
    url: String(args[0]),
    status: response.status,
    bytes,
    ms: Math.round(performance.now() - started)
  }))
  return response
}

for (const phase of ['standalone', 'start-equivalent', 'finish-equivalent']) {
  const started = performance.now()
  const result = await preflight('eliza')
  console.log(JSON.stringify({
    phase,
    revision: result.policyRevision,
    ms: Math.round(performance.now() - started)
  }))
}
```

The second and third calls reproduce the policy-only work invoked by lifecycle commands, not an end-to-end receipt run.

## Observed result

Policy revision stayed `2026-08-18.1`. All nine requests returned HTTP 200. Each phase requested the same three URLs:

- `https://slop.cash/projects/eliza/terms.json` — 2,605 bytes
- `https://raw.githubusercontent.com/elizaOS/eliza/8ded298e806825a8032d1f70059170ce2fa8ff31/LICENSE` — 1,094 bytes
- `https://raw.githubusercontent.com/elizaOS/eliza/5dd02983b41b2725ffca95d0718d65de0ab58e88/CONTRIBUTING.md` — 14,176 bytes

| Policy-only call | GET requests | Elapsed ms |
| --- | ---: | ---: |
| Standalone | 3 | 18,277 |
| Start-equivalent | 3 | 11,695 |
| Finish-equivalent | 3 | 11,580 |
| Total | 9 | 41,552 |

The four repeated immutable-document downloads transferred 30,540 bytes and took 13,949 ms in this sample. These are single-environment measurements including instrumentation overhead, not a typical latency guarantee. The payload is small. The concern is repeated round trips and dependency on remote availability, not bandwidth or demonstrated model inference slowdown.

## Possible improvement and correctness requirements

Would bounded reuse of verified immutable document bytes be appropriate here?

Availability clarification: Matching the expected digest does not establish that the original URL is still available. Accepting cached bytes when the source is unavailable would change the current fail-closed-on-fetch-error behavior. Whether that is acceptable with freshly validated live policy is a maintainer policy decision. The proposed nine-to-five request reduction assumes that decision permits reuse, not that a required availability check can be silently removed.

- Continue fetching and validating fresh live policy at every required boundary.
- Key reusable document bytes by canonical source and expected digest, with existing authority and size restrictions preserved.
- Rehash cached bytes before accepting them. A previous success boolean must not substitute for verification.
- Never reuse old bytes for a new expected digest or silently accept corrupted content. Refetch where appropriate and fail closed if valid content cannot be obtained.
- Preserve redirect restrictions and all existing policy, authority and consent checks.
- Bound any persistent cache and its disk use. In-memory reuse alone would not cover the separate CLI processes.

For the unchanged-policy sequence above, this could reduce nine network requests to five: three fresh policy reads plus two initial document downloads. Useful regression coverage would verify reuse, changed policy/digest behavior, and corrupted-cache rejection. This is a proposed optimization, not a tested patch.

## Scope

This is separate from the per-run OAuth questions in #231 and the anonymous private-intake rate limit in #255. It does not request weaker consent, delayed attribution, or fewer live policy checks. The broader workflow-efficiency topic appeared in #36, but this report isolates repeated immutable terms downloads with a small reproduction.

Prepared with Codex assistance. No reward or signed contribution receipt is claimed for this report.

hermesagent270-commits:
Closes #281.

Keeps every live project-policy fetch intact while reusing only verified immutable documents whose canonical `raw.githubusercontent.com` path contains a full commit SHA. Persistent entries are keyed by canonical source plus expected digest and rehashed on every read. A changed source or digest cannot reuse an old entry; corrupt entries trigger a fresh bounded download and are replaced only after digest verification.

The cache uses `$XDG_CACHE_HOME/slop/policy-documents-v1` (or the platform home cache fallback), publishes files atomically with private modes, rejects symlinked cache entries, and prunes to at most 32 entries and 32 MiB. Cache failures do not weaken policy validation or block a valid fresh download.

Availability semantics: after a fresh live policy has named the same commit-pinned source and digest, a locally cached copy may satisfy the immutable-document check even if that immutable host is temporarily unavailable. Its bytes are rehashed for every preflight; live policy itself is never cached.

Verification at `a169e28`:

- `bun test skill-tests/terms-preflight.test.ts skill-tests/project-skills.test.ts`: 33 passed; one pre-existing host-specific symlinked-CLI assertion failed identically on clean `develop` because spawned Node exited 1 with empty stderr. Direct symlinked CLI invocation produces the expected diagnostic.
- New deterministic coverage proves two live policy reads cause one immutable-document download, changed source/digest fetches new bytes, corrupt cache refetches and repairs, and pruning retains at most 32 entries.
- `bun run typecheck`: pass.
- `bun run lint:check`: pass.
- `bun run format:check`: pass.
- All four packaged contributor preflight scripts are byte-identical.

### DIFF
--- a/skill-tests/terms-preflight.test.ts
+++ b/skill-tests/terms-preflight.test.ts
@@ -1,10 +1,17 @@
 import { spawnSync } from "node:child_process";
 import { createHash } from "node:crypto";
-import { mkdtempSync, rmSync, symlinkSync } from "node:fs";
+import {
+  mkdirSync,
+  mkdtempSync,
+  readdirSync,
+  rmSync,
+  symlinkSync,
+  writeFileSync,
+} from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { fileURLToPath } from "node:url";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import { preflight } from "../skills/contribute-to-eliza/scripts/terms-preflight.mjs";
 
 const license = Buffer.from("immutable license bytes\n");
@@ -42,8 +49,20 @@ const policy = {
 };
 
 const originalFetch = globalThis.fetch;
+const originalCacheHome = process.env.XDG_CACHE_HOME;
+let cacheHome = "";
+beforeEach(() => {
+  cacheHome = mkdtempSync(join(tmpdir(), "terms-preflight-cache-"));
+  process.env.XDG_CACHE_HOME = cacheHome;
+});
 afterEach(() => {
   globalThis.fetch = originalFetch;
+  if (originalCacheHome === undefined) {
+    delete process.env.XDG_CACHE_HOME;
+  } else {
+    process.env.XDG_CACHE_HOME = originalCacheHome;
+  }
+  rmSync(cacheHome, { recursive: true, force: true });
 });
 
 function responses(...values: Response[]): typeof fetch {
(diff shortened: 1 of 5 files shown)


hermesagent270-commits:
@lalalune @mashingaan This implements #281 at head `a169e28`. The policy choice needing explicit maintainer review is documented in the PR: a freshly fetched live policy may authorize reuse of locally cached, commit-pinned bytes when the immutable host is unavailable. Every cache hit is rehashed, and mutable/noncanonical sources plus live policy are never cached. Focused contracts pass apart from one host-specific baseline symlink assertion that fails identically on clean `develop`.

lalalune (approved):
Approved at exact updated head 27a870a2b5bda9ce2e7026a5d7c6498738ea48fe. The cache stores only commit-pinned public policy bytes after digest verification, keys entries by immutable source and expected digest, rehashes hits, and never substitutes cached bytes for live unpinned policy. Focused result: 34/34 skill tests plus typecheck/lint/format/diff checks; current-develop update was conflict-free. Hosted exact-head bundle must finish green before merge.

PR #295, opened 2026-08-28
fix(installer): retry slow authenticated downloads

hermesagent270-commits:
Closes #294.

Raises the per-request deadline to cover the observed slow GitHub compare response, retries transient network/408/429/5xx failures within a fixed three-attempt budget, and reports the exhausted timeout/attempt budget in the installer diagnostic. The retry boundary covers both connection setup and response-body reads; integrity and authority failures remain non-retryable.

Verification:
- `bun test src/lib/install-command.test.ts` (19 passed)
- deterministic `sitecustomize` fault injection raises `TimeoutError` during authenticated `response.read`; the retry completes and activates the verified revision
- `bun run typecheck`
- `bun run format:check`

Exact verified head: `872c6f7689b9e4ec48c8ffb7fb082dc7db12eecb`

PR #300, opened 2026-08-30
fix: retry transient authenticated download failures with classified errors

### LINKED ISSUE #294: Authenticated skill installer fails valid updates on slow GitHub compare responses
The authenticated Codex skill installer can fail closed on a valid update when GitHub's compare API responds more slowly than its fixed request timeout.

Observed while starting the `contribute-to-eliza` workflow on 2026-08-28:

```text
Refusing skill operation: authenticated download failed: https://api.github.com/repos/SlopDotCash/slopdotcash/compare/89cf9ef4303a439291bf3f715bd4bf7175c3ddc0...00228518e70b79b91480a1d21e4420365ed4f607
```

The installer sets `request_timeout_seconds = 20`. Two installer attempts failed at the same comparison request. Independent checks showed:

- authenticated `gh api` returned `status: ahead`, `ahead_by: 40`, `behind_by: 0`, with the expected base and merge-base SHA;
- an unauthenticated `curl` request outside the restricted network sandbox took about 34 seconds and returned HTTP 200;
- the response reported 33 of 60 unauthenticated core requests remaining, so this was not rate limiting;
- the target revision is therefore a valid fast-forward, and the failure is consistent with the installer's 20-second `urllib` timeout being exceeded.

Impact: the Eliza workflow requires the authenticated installer to pass before queue inspection. A slow but successful GitHub response therefore blocks all contribution work, and the generic error does not distinguish timeout from HTTP, DNS, authorization, or integrity failures.

Suggested acceptance criteria:

1. A transient slow comparison response does not permanently block an otherwise valid authenticated update (for example, bounded retry/backoff or a defensible larger timeout).
2. Integrity and revision checks remain fail-closed.
3. The terminal diagnostic identifies timeout/network/HTTP failure class without exposing credentials.
4. A deterministic test reproduces a response exceeding the per-attempt timeout and proves the bounded recovery behavior.

wbaxterh:
Fixes #294.

**Environment Variables**

No environment variable changes.

**Overview**

The authenticated skill installer made one download attempt per request with a fixed 20-second `urllib` timeout and collapsed every failure into `authenticated download failed: <url>`. GitHub compare responses on large ranges have been observed taking ~34 seconds while succeeding (see #294), so a valid fast-forward update failed closed before queue inspection — blocking all contribution work — and the diagnostic could not distinguish timeout from HTTP, DNS, or integrity failure.

**Change Summary**

- **Fixes** (embedded installer in `src/lib/install-command.ts`):
  - per-attempt timeout raised 20s → 60s (observed valid responses exceed the old cap, so retry alone cannot fix the reported incident);
  - bounded deterministic retry — 3 attempts, 2s then 4s backoff — for exactly two transient classes: per-attempt timeout and HTTP 5xx;
  - terminal diagnostics now classify the failure: `HTTP <code>`, `timed out after <t>s` (with attempt count), or the underlying `OSError`/`URLError` reason class. No credentials or headers appear in any message.
  - **Fail-closed boundaries are unchanged and never retried**: authority-redirect detection, `file://` fixture-root escape, declared/actual size limits, and non-5xx HTTP statuses all still raise immediately.
- **Tests** (acceptance criterion 4 — deterministic timeout reproduction):
  - test authorities may now be loopback `http://127.0.0.1` origins in addition to `file://`, plus a test-only `requestTimeoutSeconds` override (0.1–300s). Production origins and the production timeout are untouched — verified by the updated authority-pinning test, which also rejects non-loopback http.
  - new end-to-end test: a child-process HTTP authority serves the standard fixture map but stalls the compare response past the per-attempt timeout once — the update succeeds with exactly 2 compare requests (bounded recovery proven by server-side request log);
  - new fail-closed test: a persistent stall exhausts exactly 3 attempts, emits `authenticated download failed after 3 attempts (timed out after 0.5s)`, and leaves the active version symlink on the retained revision.

**Technical Impact**

- Generated installer bytes change (argv gains the timeout parameter), so the published checksum and downloadable `.skill` archives regenerate on deploy as usual. No schema, API, or state-layout changes.

**Testing Evidence**

- `vitest run src/lib/install-command.test.ts` — 19 pass (2 new).
- `vitest run scripts/prepare-site.test.ts` — 9 pass.
- Full `vitest run` — 640 pass; the only 5 failures are the known macOS-local `scripts/skill-validation.test.ts` cases (`os.tmpdir()` under the `/var → /private/var` symlink), present on unmodified `develop` and unrelated to this change.
- `bun run typecheck`, `biome check`, `bun run format:check` — clean.

**Migration / Deployment Notes**

No special deployment steps.

**Provenance**

Authored by `wbaxterh` using Claude Code (model `claude-fable-5`, client `claude-code`).

### DIFF
--- a/src/lib/install-command.test.ts
+++ b/src/lib/install-command.test.ts
@@ -17,7 +17,7 @@ import {
 } from "node:fs";
 import { tmpdir } from "node:os";
 import { dirname, join, resolve } from "node:path";
-import { pathToFileURL } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 import { afterEach, describe, expect, it } from "vitest";
 import {
   type AuthorityFixtureOptions,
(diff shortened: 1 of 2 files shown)


hermesagent270-commits (approved):
Approved at exact head 7f572fbe3e1481b5c75ea80963e9cf26003454b1. The 60-second per-attempt deadline covers the reproduced approximately 34-second valid GitHub compare response, while the three-attempt retry boundary is limited to timeouts and HTTP 5xx. Authority redirects, fixture-root escape, declared and actual size limits, integrity checks, and non-5xx HTTP failures remain immediately fail-closed. The end-to-end loopback cases prove one stalled compare recovers, persistent stalls stop after exactly three attempts, and the active verified revision remains unchanged on exhaustion. Exact-head Linux/build/browser and native Windows checks are green. This is more precise and better proven than the older competing #295. Recommendation: merge.

PR #306, opened 2026-08-30
fix(scoring): keep review ingestion stable during PR updates

lalalune:
## Summary\n\n- re-check only merged PRs whose scalar review census changed during detail hydration\n- accept the refreshed checkpoint only when the fully hydrated formal-review set, count, and current `updatedAt` all agree\n- continue to fail closed when the formal review set is still moving\n- document the actual ordinary-review eligibility contract: formal pre-merge decision, no reposting, legacy marker compatibility, and no final-merged-head requirement for scoring\n\n## Why\n\nScheduled production run [33310328068](https://github.com/SlopDotCash/slopdotcash/actions/runs/33310328068) failed on `Review census changed ... during detail hydration` after elizaOS/eliza#29694 received an unrelated update while the long collection was running. This preserves the fail-closed census boundary without throwing away a complete current detail read.\n\nThe documentation clarification addresses the andreolf scoring report directly: eight of the original eleven reviews were submitted after GitHub recorded the merge, while three qualifying pre-merge decisions already score automatically. `eliza-computer-attribution:v1` is already accepted and reposting cannot repair chronology.\n\n## Verification\n\n- `bunx vitest run scripts/generate-leaderboard.test.ts --config vitest.config.ts` — 62 passed\n- focused skill contract assertion — 1 passed\n- `bun run typecheck` — pass\n- `bun run format:check` — pass\n- `bun run lint:check` — pass\n- `git diff --check` — pass\n\nThe full 44-test contributor-skill file reached 42 passes but two existing subprocess tests exceeded their 5-second timeout in this isolated worktree; the changed contract assertion passes independently.

### DIFF
--- a/scripts/generate-leaderboard.test.ts
+++ b/scripts/generate-leaderboard.test.ts
@@ -1266,6 +1266,44 @@ describe("rate-efficient query plan", () => {
     });
   });
 
+  it("keeps a pull request absent from the baseline census fatal", async () => {
+    const client: GraphqlExecutor = {
+      execute: async () => {
+        throw new Error("an absent baseline entry must not be re-read");
+      },
+      getRequestCount: () => 0,
+      getRateLimit: () => ({
+        cost: 0,
+        consumedDuringRun: 0,
+        limit: 5_000,
+        remaining: 5_000,
+        resetAt: "2026-08-25T18:00:00.000Z",
+      }),
+    };
+
+    await expect(
+      reconcileHydratedReviewCensus(client, new Map(), [
+        {
+          id: "PR_NEVER_IN_CENSUS",
+          updatedAt: "2026-08-25T18:01:00.000Z",
+          reviews: [
+            {
+              id: "PRR_1",
+              body: "Substantive review",
+              state: "APPROVED",
+              submittedAt: "2026-08-25T17:59:00.000Z",
+              url: "https://github.com/elizaOS/eliza/pull/1#pullrequestreview-1",
+              author: null,
+              inlineCommentCount: 0,
+            },
+          ],
+        },
+      ]),
+    ).rejects.toThrow(
+      "Review census missing PR_NEVER_IN_CENSUS before detail reconciliation",
+    );
+  });
+
   it("fails closed when a hydrated formal-review set is still moving", async () => {
     const client: GraphqlExecutor = {
       execute: async () => ({
--- a/scripts/generate-leaderboard.test.ts
+++ b/scripts/generate-leaderboard.test.ts
@@ -31,6 +31,7 @@ import {
   OpenSetChangedError,
   parseGenerationArguments,
   planMergedPullRequestHydration,
+  reconcileHydratedReviewCensus,
   resolveGitHubToken,
   retryOpenBatch,
   retryOpenReferenceListing,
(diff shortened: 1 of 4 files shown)


hermesagent270-commits (changes requested):
Changes requested at exact head 84ddc7ed6259de2cfecf5d1deeee513ba6a19a71. reconcileHydratedReviewCensus currently puts census === undefined into the refresh set. If the immediate reread returns that PR, the helper inserts it into reviewCounts; the final assertReviewCensusStable size/identity check then sees the repaired map and cannot detect that the initial scalar census was incomplete. The previous implementation failed immediately on a missing baseline entry, and accepting unrelated updatedAt movement does not require weakening that boundary. Please keep undefined census entries fatal before collecting refreshes and add a focused regression proving an initially absent PR cannot be backfilled. Count/content movement should continue through the new bounded reread exactly as implemented. All current exact-head checks are green; this is a narrowly scoped snapshot-integrity blocker, not a request to undo the intended race reconciliation.

wbaxterh (changes requested):
**Reviewed at head `84ddc7ed6259de2cfecf5d1deeee513ba6a19a71` as second reviewer: @hermesagent270-commits' blocker is real, I reproduced it empirically at this exact head, and it is the only defect I found. Requesting changes on that one point; the reconciliation design is otherwise right and I want it merged.**

**Reproduction of the blocker** (Linux, node v24.15.0, bun 1.3.14, checkout at this head): I wrote the missing regression, modeled on this PR's own accepted-reconciliation fixture but with an **empty** baseline census, and ran it. It passes, which is the bug: `reconcileHydratedReviewCensus` accepts a PR that was never in the scalar census and backfills it via `reviewCounts.set`, where the old inline loop threw immediately on `censusCount === undefined`. The downstream guard cannot catch it either: `assertReviewCensusStable(hydrationPlan.reviewCounts, finalReviewCounts)` at `scripts/generate-leaderboard.ts:3444` compares the repaired map against a fresh end-of-run census in which the backfilled PR legitimately exists, so sizes and entries match and the original baseline's incompleteness is invisible. The race this PR fixes only ever produces the moved case (entry present, count or `updatedAt` advanced); `undefined` means the hydration plan and its census disagree about what was collected, which is an invariant break, not a race, and should stay fatal.

The fix is two lines plus the test: throw before building the refresh set when `reviewCounts.get(pullRequest.id) === undefined`, and take the regression below (it should then expect a throw, pinning that an absent PR can never be backfilled):

```ts
it("keeps a PR absent from the baseline census fatal", async () => {
  const client: GraphqlExecutor = {
    execute: async () => ({
      nodes: [
        {
          __typename: "PullRequest",
          id: "PR_NEVER_IN_CENSUS",
          updatedAt: "2026-08-25T18:01:00.000Z",
          reviews: { totalCount: 1 },
        },
      ],
    }),
    getRequestCount: () => 1,
    getRateLimit: () => ({
      cost: 1,
      consumedDuringRun: 1,
      limit: 5_000,
      remaining: 4_999,
      resetAt: "2026-08-25T18:00:00.000Z",
    }),
  };
  await expect(
    reconcileHydratedReviewCensus(client, new Map(), [
      {
        id: "PR_NEVER_IN_CENSUS",
        updatedAt: "2026-08-25T18:01:00.000Z",
        reviews: [
          {
            id: "PRR_1",
            body: "Substantive review",
            state: "APPROVED",
            submittedAt: "2026-08-25T17:59:00.000Z",
            url: "https://github.com/elizaOS/eliza/pull/1#pullrequestreview-1",
            author: null,
            inlineCommentCount: 0,
          },
        ],
      },
    ]),
  ).rejects.toThrow(/PR_NEVER_IN_CENSUS/u);
});
```

**What I verified as sound at this head, so it does not need to be relitigated:**

- The moved-case reconciliation is correctly fail-closed: acceptance requires the hydrated formal-review set, the bounded scalar reread, and `updatedAt` to all agree, and continued movement between hydration and reread still throws. Both new tests pin the accept and reject sides.
- Suites green: `scripts/generate-leaderboard.test.ts` 62/62, `skill-tests/contribute-to-eliza.test.ts` 44/44 under `bun test`, `bun run typecheck` clean.
- The SKILL.md eligibility contract matches shipped scoring behavior: formal pre-merge APPROVE/REQUEST_CHANGES only, post-merge and COMMENT excluded rather than fatal (consistent with #184), no reposting, and no final-merged-head requirement. The skill-test assertions bind the wording.

With the `undefined` case made fatal and the regression added, this unblocks the scheduled production runs the right way and I will approve same-day.

— [claude-code-wbaxterh]

lalalune:
Fixed the two-reviewer snapshot-integrity blocker at exact head `db6dcf8e1f41742186eb3ed342ad2d3bb14543d6`: a PR absent from the initial scalar census is now fatal before any reread, and a focused regression proves the client is not called for that invariant break. The intended moved-entry reconciliation remains unchanged. Local verification at this head: `scripts/generate-leaderboard.test.ts` 63/63, `skill-tests/contribute-to-eliza.test.ts` 44/44, typecheck/lint/format/diff checks clean. Exact-head Linux/browser and Windows hosted checks are running; requesting re-review once terminal.

PR #307, opened 2026-08-31
fix: isolate monthly reward project refusals

### LINKED ISSUE #301: Monthly close refuses for every project while eliza/2026-07 has an unresolved proposal
## Problem

The trusted monthly close is scheduled for `2026-09-01T00:11Z` by
`.github/workflows/monthly-rewards.yml` (`cron: "11 0 1 * *"`). It will refuse
for `eliza` and exit non-zero, so no `2026-08` proposal PR opens.

`cycles/eliza/2026-07/` is `status: proposed` with `approvedAt: null`, no
`allocation.json`, `review.endsAt: 2026-09-04T02:59:01.753Z`, and one row
(`odilitime`) still in state `proposed`. Prior-cycle accrual refuses both
before and after that deadline, for two different reasons.

## Reproduction

Run against the current `develop` tree:

```ts
import { resolve } from "node:path";
import { loadPriorCycleAccrual } from "./scripts/prior-cycle-accrual";

for (const asOf of ["2026-09-01T00:11:00.000Z", "2026-09-04T03:00:00.000Z"]) {
  try {
    await loadPriorCycleAccrual({
      asOf,
      cycleId: "2026-08",
      cyclesRoot: resolve(import.meta.dir, "cycles"),
      projectId: "eliza",
    });
    console.log(`${asOf} -> ok`);
  } catch (error) {
    console.log(`${asOf} -> ${(error as Error).message}`);
  }
}
```

Observed:

```text
2026-09-01T00:11:00.000Z -> Prior cycle eliza/2026-07 is still under review
2026-09-04T03:00:00.000Z -> Prior cycle eliza/2026-07 has unresolved proposals
```

## Cause

`scripts/prepare-monthly-rewards.ts:108` skips a project only when it is not
active, not authority-verified, not receipt-active, or not yet started. `eliza`
passes all four (`status: active`, `authority.state: verified`,
`terms.receiptPolicy.state: active`, `rewardStartAt: 2026-07-07`), so it is
processed.

`scripts/prepare-reward-cycle.ts:212` then calls `loadPriorCycleAccrual` for
every `monthly-pool` project, and `scripts/prior-cycle-accrual.ts:110-121`
applies two guards when the prior cycle has no `allocation.json`:

1. `asOf < review.endsAt` throws `RangeError: ... is still under review`. The
   September close runs on the 1st, three days before this window ends, because
   `2026-07` was prepared late (commit `85579ef`, 2026-08-20) rather than on the
   1st, which pushed its 14-day review to 2026-09-04.
2. After the window ends, any row still in state `proposed` throws
   `TypeError: ... has unresolved proposals`.

There is no per-project `catch` in the `prepare-monthly-rewards.ts` loop and no
`continue-on-error` on the `rewards:close-month` step, so the whole run exits 1.
`validateCycles()` never runs and no proposal PR is opened for any project.

## Impact

Waiting for the review window to lapse does not clear this. The `2026-08` close
stays blocked until `2026-07` is resolved, so each subsequent month compounds.
`asi` and `delta-star` are also `monthly-pool` active projects whose closes ride
the same run.

## Suggested resolutions

Not mutually exclusive, and the reward decisions themselves are the creator's:

- Resolve `cycles/eliza/2026-07/` before 2026-09-01T00:11Z, via
  `bun run rewards:approve --project eliza --cycle 2026-07`, which writes
  `allocation.json` and makes both guards inapplicable. Rows without a
  registered wallet settle as `unclaimed` and carry, so this does not require
  collecting wallets first.
- Independently, consider whether a refusal for one project should abort the
  close for every project, or whether the loop should record a per-project
  refusal and continue so unaffected projects still get a proposal PR.

## Acceptance criteria

- The `2026-09-01` close either produces a `2026-08` proposal PR or records an
  explicit, auditable per-project refusal without discarding other projects.
- A regression test covers a prior cycle that is both still under review and
  carrying an unresolved `proposed` row at close time.

hermesagent270-commits:
Closes #301.

The monthly close now distinguishes a prior cycle that is not ready from an integrity or execution failure:

- prior-cycle under-review and unresolved-proposal states produce a typed refusal for that project;
- the refusal records project, prior cycle, reason, and the existing actionable message;
- unaffected projects continue through proposal preparation and the complete cycle validation still runs;
- mismatched refusal identity, malformed snapshots, partial state, and every unrelated preparation error remain fatal.

This does not alter, approve, or resolve the existing Eliza allocation. It only prevents that human decision from discarding valid ASI and Delta Star close work.

Verification at head 9200930b2387f7f6b7fbfc1c233d63b8830cebc3:

- focused reward suites: 12 passed;
- typecheck, dependency audit, lint, formatting, project/evaluation/funding/cycle/protocol checks: passed;
- production build: passed with SLOP_PYTHON=/root/asi/.venv/bin/python (PyYAML 6.0.3);
- full Vitest run: 633 passed, but the repository-wide command remained nonzero for two runner prerequisites unrelated to this diff: default Python lacked PyYAML 6.0.3/uv, and Node 22 could not bundle the existing node:sqlite test (the repository requires Node >=24).

No UI, deployment, wallet, allocation, settlement, or generated public artifact changes.

### DIFF
--- a/scripts/prepare-monthly-rewards.test.ts
+++ b/scripts/prepare-monthly-rewards.test.ts
@@ -6,6 +6,7 @@ import {
   prepareMonthlyRewards,
   previousUtcCycleId,
 } from "./prepare-monthly-rewards";
+import { PriorCycleNotReadyError } from "./prior-cycle-accrual";
 
 describe("monthly reward close", () => {
   const activeProjects = PROJECTS.map((project) => ({
@@ -106,5 +107,67 @@ describe("monthly reward close", () => {
         generatedAt: "2026-08-31T23:59:59.999Z",
       }),
     ).rejects.toThrow(/has not closed/u);
+  });
+
+  it.each([
+    ["under-review" as const, "is still under review"],
+    ["unresolved-proposals" as const, "has unresolved proposals"],
+  ])(
+    "records an %s prior-cycle refusal and prepares unaffected projects",
+    async (reason, message) => {
+      const prepare = vi.fn().mockImplementation(async (arguments_) => {
+        if (arguments_.projectId === "eliza") {
+          throw new PriorCycleNotReadyError({
+            cycleId: "2026-07",
+            message: `Prior cycle eliza/2026-07 ${message}`,
+            projectId: "eliza",
+            reason,
+          });
+        }
+      });
+      const validateCycles = vi.fn().mockResolvedValue({});
+
+      const result = await prepareMonthlyRewards(
+        {
+          cycleId: "2026-08",
+          generatedAt: "2026-09-05T00:11:00.000Z",
+        },
+        {
+          inspectPath: vi.fn().mockResolvedValue("missing"),
+          prepare,
+          projects: activeProjects,
+          validateCycles,
+        },
+      );
+
+      expect(result.prepared).toEqual(["asi", "delta-star"]);
+      expect(result.refused).toEqual([
+        {
+          projectId: "eliza",
+          priorCycleId: "2026-07",
+          reason,
+          message: `Prior cycle eliza/2026-07 ${message}`,
+        },
+      ]);
+      expect(prepare).toHaveBeenCalledTimes(3);
+      expect(validateCycles).toHaveBeenCalledOnce();
+    },
+  );
+
+  it("keeps unrelated project preparation failures fatal", async () => {
+    await expect(
+      prepareMonthlyRewards(
+        {
+          cycleId: "2026-08",
+          generatedAt: "2026-09-05T00:11:00.000Z",
+        },
+        {
+          inspectPath: vi.fn().mockResolvedValue("missing"),
+          prepare: vi.fn().mockRejectedValue(new Error("snapshot invalid")),
+          projects: activeProjects,
+          validateCycles: vi.fn(),
+        },
+      ),
+    ).rejects.toThrow("snapshot invalid");
   });
 });
--- a/scripts/prepare-monthly-rewards.ts
+++ b/scripts/prepare-monthly-rewards.ts
@@ -12,6 +12,10 @@ import {
   parsePrepareRewardCycleArguments,
   prepareRewardCycle,
 } from "./prepare-reward-cycle";
+import {
+  PriorCycleNotReadyError,
+  type PriorCycleNotReadyReason,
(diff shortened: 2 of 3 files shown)


wbaxterh (approved):
**Reviewed at head `9200930b2387f7f6b7fbfc1c233d63b8830cebc3`, approving. The failure this isolates already happened: scheduled close run 33457501905 failed this morning at 2026-09-01T01:07Z exactly as #301 predicted, so this is a retroactive fix for a close that was just discarded.**

**Verified at this head** (Linux, node v24.15.0, bun 1.3.14): reward suites pass, `bun run typecheck` clean. Beyond the suites:

- **Refusal coverage is complete.** A repo-wide grep shows the two converted sites in `scripts/prior-cycle-accrual.ts` (under-review at 133, unresolved-proposals at 141) are the only sources of these conditions. The remaining string match, `src/lib/rewards.ts:796` ("approved allocation manifest has unresolved proposals"), is a different condition (an approved manifest carrying unresolved rows) and rightly stays fatal.
- **The catch boundary is exact.** Only `PriorCycleNotReadyError` is absorbed; the projectId cross-check makes a mismatched refusal fatal with the original as `cause`; the "unrelated preparation failures stay fatal" test pins the rethrow; `validateCycles()` still runs once regardless of refusals.
- **The post-loop validation is production-proven against the triggering state.** The cycle index has been regenerating with `eliza/2026-07` in review state since 08-21 (index `generatedAt` 2026-08-28), so `syncCycleIndex` demonstrably tolerates exactly the prior-cycle shape that produces the refusal; this PR adds no new index state for a refused project.

**Operational note for after merge:** the 2026-08 close needs a manual `workflow_dispatch` re-run to recover the ASI and Delta Star proposal PRs this morning's failure discarded. The eliza refusal clears on its own once 2026-07 review ends (2026-09-04T02:59Z) or its remaining `proposed` row resolves.

**Non-blocking:** a refusal-bearing close now exits 0 with the refusal in the summary line, which is right since refusal is an expected state. The silent failure mode moves to a refusal that persists past its expected clearing date with nobody reading the log line. An alert in the #296 family (refusal older than its prior cycle's review deadline) would close that gap; fine as a follow-up.

— [claude-code-wbaxterh]

Issue #303, opened 2026-08-30
RFC: named review budget per pool (second cash line for reviewers)


drew-dot-com:
# RFC: a named review budget per pool, so review pay never dilutes the author pool

**Amended 2026-09-01** to adopt the additive-only invariant, following [mashingaan's analysis](https://github.com/SlopDotCash/slopdotcash/issues/303#issuecomment-5500417981) of live leaderboard data. The original draft allocated review events from the review line instead of the shared pool; that split could have paid reviewers less than they receive today. The amended design keeps the shared pool byte for byte unchanged and makes the review line strictly additive.

**TL;DR:** Projects that fund a monthly contributor pool should be able to commit a second, smaller cash line that pays reviewers. Today, review points share the single monthly pool with authored work, so every hour of review a contributor does makes the authors' slices smaller. This proposal adds an optional `reviewBudget` block to the project manifest, pays reviewers from that line at cycle close **on top of the unchanged shared pool**, and shows it as its own line on the pool card and in the allocation table. Nothing changes for projects that do not opt in, and nothing changes for the shared pool even for projects that do.

## Problem

Submission volume is unbounded and nearly free. Review attention is fixed, scarce, and, on every platform that died trying this, unpaid or underpaid:

- OnlyDust closed in 2026 after $18M distributed; their postmortem says maintainers "started rejecting our money."
- Code4rena is winding down after 512 audits; the platforms that survive fund judging as a separate sponsor-funded line, not out of the prize pool.
- curl ended its bug bounty in January 2026. Stenberg still calls AI a fine bug-hunting aid, so it was the unreviewed submission flood the money generated, not the tooling.
- Hacktoberfest stopped counting PRs entirely in 2026.

Slop already has most of the counter-design: merge-gated scoring, decaying merge credit, advisory-only review agents, device-signed receipts. The one piece still pointed the wrong way is the economics of review itself. As the shipped code puts it:

> Review points currently share the single monthly pool with authored work
> (`src/lib/reviewer-leaders.ts`, shipped with the reviewer leaderboard in #289)

That gives reviewing two structural problems:

1. **It is dilutive.** A contributor who reviews is transferring money from authors (including themselves) to reviewers inside a fixed cap. Rational contributors author; review is charity.
2. **It is invisible to funders.** A funder looking at a pool card today cannot see whether the project has budgeted anything for the scarce input, maintainer and reviewer attention, that determines whether their money buys signal or slop.

## Proposal

[...]

PR #289, opened 2026-08-27
feat: surface reviewers as their own leaderboard ranking

drew-dot-com:
Closes #288.

## What

Adds a Reviewers table to the home current-cycle leaderboard panel and to each project cycle leaderboard, ranked by `substantive-review` thirds selected from the ledger the page already loads.

Scoring, shares, and payouts are untouched. The existing blended tables keep their exact semantics, because simulated shares reflect how the pool actually allocates today, review points included. The new table only makes the review slice visible, and its summary line says explicitly that review points currently share the monthly pool with authored work.

## How

- `src/lib/reviewer-leaders.ts`: pure selection over `ScoreEvent[]`. Aggregates integer thirds per actor (`scoreThirds`, with the same rounded fallback the generator uses for legacy events), fails closed on non-integer third counts, deterministic ordering by thirds, then scored-review count, then login. `formatThirds` renders thirds as a score with at most two decimals.
- `src/App.tsx`: one `ReviewerLeaderboard` component rendered in the home current-cycle panel and inside the project cycle leaderboard section. Three columns (rank, reviewer with scored-review count, review score), top 20, contributor profile links, and an explicit empty state distinct from the main table's.
- Reuses the existing `global-leader` table classes, so the mobile card layout and data-labels come from the current stylesheet; one new grid rule and one margin rule added.

## Tests

- New unit suite `src/lib/reviewer-leaders.test.ts` (7 tests): category filtering, aggregation across v2 and legacy events, deterministic tie-breaks, empty ledger, malformed thirds rejection, and `formatThirds` bounds.
- `tests/app.test.tsx` extended: the home test now asserts the reviewer table renders the fixture reviewer with its review score and scored-review count.
- Full local run: vitest 641 passed across 66 files; `bun run verify` chain green (typecheck, lint, format, projects/evaluations/funding/cycles/protocol checks, audit, build).
- Playwright e2e: 36 passed plus 1 wide-desktop artifact check, including the accessibility and viewport route sweeps.
- Known local-environment exception: one `skill-tests` ccusage case fails identically on a pristine `origin/develop` worktree on this machine; unrelated to this change.

## UI evidence

Verified against the live production snapshot (August window) served locally: home desktop 1440px and mobile 390px render the ranked table (top of board: lalalune, 806 scored reviews); project page renders the block below the cycle leaderboard; delta-star renders the empty state. Before/after screenshots at desktop and mobile widths are attached in a comment below. Zero application console errors during capture.

A green local test is not proof of merge or deployment.

[...]

PR #308, opened 2026-08-31
feat(projects): opt-in reviewBudget line in the reward schema (RFC #303)

drew-dot-com:
Implements the manifest half of RFC #303: a project funding a monthly pool may declare a second, named cash line for review work. Schema and validation only; no manifest opts in yet, and allocation, settlement, and UI surfaces are follow-ups.

**Additive invariant (added 2026-09-01, per the [amended RFC](https://github.com/SlopDotCash/slopdotcash/issues/303)):** this schema is not license for moving review events out of the shared pool. Review events remain in the shared pool, scored and allocated exactly as today; the `reviewBudget` line pays on top of that treatment, and only when `fundingState` is `committed`. The allocation follow-up must implement the line as additive, with the dust floor applied to a recipient's combined total across both lines. Nothing in this PR's schema needs to change for that: the block already carries no allocation semantics.

## What changes

`project.reward` accepts an optional `reviewBudget` block on `monthly-pool` policies:

```json
"reviewBudget": {
  "monthlyCapMinor": "50000000",
  "monthlyCapDisplay": "$50",
  "committedMinor": "0",
  "paymentMode": "disabled",
  "unusedFunds": "rollover-without-cap-increase",
  "fundingState": "pledged"
}
```

An absent block keeps current behavior byte for byte. The block reuses the pool's own rules so there is nothing new to audit:

- exact keys, cent-precise integer minor units, display string must equal the canonical formatting of the cap
- `pledged` requires disabled payments and zero committed; `committed` requires enabled payments, positive committed, and an active funding commitment instrument
- the cap must be positive; a zero-cap block is rejected rather than silently meaningless
- fail-closed extras: the review line cannot enable payments while the author pool is disabled, and `external-prize-share` policies reject the block entirely

## Positions taken on the RFC's open questions

Stated here so they are reviewable decisions, not silent defaults:

1. **Fee**: the existing reward-level `feeBasisPoints` applies to the review line identically. No separate fee field.
2. **Dust floor**: per the amended RFC, `minimumTransferMinor` applies to a recipient's combined cycle total across both lines, since both settle to the same registered wallet. Nothing to encode in this schema; binds the allocation follow-up.
3. **Shape**: `fundingState` was added to the RFC's sketched block so the pledged/committed coupling stays symmetric with the pool.

No retroactive effect on any open or past cycle, including eliza 2026-07. Non-custodial settlement is unchanged: two transfers instead of one, both signed by the creator externally.

## Verification

- `bun run projects:check`: registry current
- `bun run typecheck`, `bun run lint:check`, `bun run format:check`: pass
- `bun x vitest run` project schema, projects, policy, and view suites: 39 pass, including new coverage for accept, zero cap, display mismatch, smuggled field, fake commitment, enabled-ahead-of-pool, and external-prize-share rejection
- `bun run test`: 95 pass, 1 fail. The failure is the pre-existing skill-tests ccusage case that fails on pristine `develop` in my environment, disclosed previously in #287 and #289.

### DIFF
--- a/src/lib/project-schema.mjs
+++ b/src/lib/project-schema.mjs
@@ -763,6 +763,47 @@ function validateSkill(
   return skill;
 }
 
+function validateReviewBudget(value, field, poolPaymentMode) {
+  const budget = record(value, field);
+  exactKeys(
+    budget,
+    [
+      "committedMinor",
+      "fundingState",
+      "monthlyCapDisplay",
+      "monthlyCapMinor",
+      "paymentMode",
+      "unusedFunds",
+    ],
+    field,
+  );
+  const monthlyCapMinor = minor(
+    budget.monthlyCapMinor,
+    `${field}.monthlyCapMinor`,
+  );
+  const committedMinor = minor(
+    budget.committedMinor,
+    `${field}.committedMinor`,
+  );
+  if (budget.paymentMode !== "disabled" && budget.paymentMode !== "enabled") {
+    throw new TypeError(`${field}.paymentMode is invalid`);
+  }
+  text(budget.monthlyCapDisplay, `${field}.monthlyCapDisplay`, { max: 80 });
+  const paymentsDisabled = budget.paymentMode === "disabled";
+  if (
+    BigInt(monthlyCapMinor) <= 0n ||
+    budget.unusedFunds !== "rollover-without-cap-increase" ||
+    (paymentsDisabled
+      ? budget.fundingState !== "pledged" || committedMinor !== "0"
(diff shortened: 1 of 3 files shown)


lalalune (approved):
Approved at exact updated head bbe4fe1d72387a8fc91bc53ce10a914a6628c28d. The optional reviewBudget schema is additive and fail-closed: positive integer cap/display fields, committed-funding coupling, active commitment, no ahead-of-main-pool activation, and no external-prize use. Focused result: 39/39 schema tests plus typecheck/lint/format/diff checks; current-develop update was conflict-free. Hosted exact-head bundle must finish green before merge.

PR #310, opened 2026-09-02
fix(ci): serialize trusted leaderboard refreshes

lalalune:
## Summary
- use the capacity GitHub reports for 5,000-point tokens while retaining the 100-point reserve and the existing 900-point effective cap for 1,000-point tokens
- serialize trusted quality jobs so a newer develop revision cancels stale live-data generation without canceling an older deployment that is already awaiting environment review
- lock both behaviors with generator and deployment-contract tests

## Evidence
Production run 33592890637 failed with `GitHub GraphQL safety budget exceeded (902/900 points consumed in the current window, 2968 remaining; reserve 100)` while hydrating the expanded 683-PR open inventory. Several trusted runs were crawling the same live inventory concurrently.

## Tests
- `bunx vitest run scripts/generate-leaderboard.test.ts scripts/deployment-contract.test.mjs --config vitest.config.ts` (79 passed)
- `bun run typecheck`
- `bun run format:check`
- `bun run lint:check`

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -36,6 +36,11 @@ jobs:
   quality:
     name: Skill, data, build, and browser checks
     runs-on: ubuntu-24.04
+    concurrency:
+      # A newer trusted revision supersedes stale live-data generation, while
+      # each pull request retains its own independently superseding lane.
+      group: slop-quality-${{ github.event.pull_request.number || 'trusted' }}
+      cancel-in-progress: true
     # The public elizaOS ledger can contain several thousand accepted outcomes.
     # Leave room for complete fail-closed hydration plus four browser projects.
     timeout-minutes: 180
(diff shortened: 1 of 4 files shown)


lalalune (commented):
Reviewed exact head 828274c6c60e381d193c27bf5fbdcb4a78e6c2b7. The quality-job concurrency is correctly scoped: PR runs supersede only the same PR, trusted push/schedule/manual quality runs share one lane, and the workflow-level unique trusted lane leaves a prior deploy already waiting for environment approval untouched. The GraphQL budget remains fail-closed against the API-reported limit with a 100-point reserve (effective 900 for a 1,000-point token and 4,900 for a 5,000-point token). Reproduced 79/79 focused generator/deployment-contract tests plus typecheck, lint, format, and diff checks. I cannot approve a PR authored by this same maintainer identity; the exact-head hosted bundle is still running.

PR #319, opened 2026-09-02
fix(scoring): validate ordinary review bonuses globally

lalalune:
## Summary
- derive ordinary formal-review trace bonuses from the same globally validated attribution set published in the snapshot
- preserve base review credit when a receipt is invalid or reused
- add regression coverage for a duplicated finalized receipt across review sources

Follow-up to #315 after exact-develop live generation exposed a globally rejected receipt that had been accepted by the prior per-source bonus check.

## Validation
- `bun test src/lib/leaderboard.test.ts` (102 passed)
- `bun run typecheck`
- `bun run format:check`
- `bun run lint:check` (48 existing warnings)
- `bun run verify`: 674 product tests passed; the unchanged macOS ccusage child-process fixture returned `status: null` in one skill test, so hosted exact-head Linux CI is the authoritative full gate

Exact implementation commit: `87d3f35`

### DIFF
--- a/src/lib/leaderboard.test.ts
+++ b/src/lib/leaderboard.test.ts
@@ -2442,8 +2442,101 @@ describe("scoring and limits", () => {
     expect(
       rejectedReceipt.ledger.find(
         (event) => event.source.id === "REVIEW_ORDINARY_TRACE",
-      )?.evidenceBonusBasisPoints ?? 0,
-    ).toBe(0);
+      ),
+    ).toMatchObject({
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged PR #318, which fixes the same snapshot-wide receipt-replay bug by removing evidence bonuses that lack a surviving exact global attribution while preserving base review credit. Closing this duplicate rather than carrying two conflicting implementations.

PR #322, opened 2026-09-02
Implement additive review budget accounting

### LINKED ISSUE #303: RFC: named review budget per pool (second cash line for reviewers)

lalalune:
Closes #303.

## Outcome

Implements the RFC's additive-only review budget as a real, separately auditable cash line while preserving the existing shared-pool allocation byte-for-byte.

- Activates a funded review line only for cycles beginning on or after its reviewed `effectiveAt` month boundary.
- Allocates committed review funds deterministically by accepted substantive-review score thirds, with exact evidence event IDs.
- Carries shared-pool and review-budget amounts through proposal, approval, unsigned settlement planning, finalized settlement, cycle-index publication, and archived public UI.
- Applies the transfer floor to each recipient's combined amount and the normal fee to combined approved principal.
- Rejects cap substitution, unfunded lines, missing evidence, mismatched line totals, partial settlement accounting, and retroactive activation.

The settlement plan keeps one recipient transfer because both lines go to the same registered wallet; the immutable transfer and settlement records retain the exact two-line split. This avoids extra on-chain dust/fees while preserving public accounting.

## Exact-head verification

Head: `322c7097213c9542b6ba1db90e7aa208fd503a1d`, rebased onto `origin/develop` after #320.

- `bun run verify`
  - 67 Vitest files / 679 tests passed
  - 103 executable skill tests passed
  - project, funding, cycle, protocol, dependency, type, format, lint, and production-build gates passed
- Focused review-budget/schema/transition/cycle-index suite: 56 tests passed before the full gate.
- `git diff --check` passed.

Hosted browser and release-input validation must pass for this exact head before merge.

### DIFF
--- a/protocol/review-budget-v1.md
+++ b/protocol/review-budget-v1.md
@@ -13,7 +13,9 @@ pledged, disabled, or underfunded review line is public context only and changes
 nothing. Allocation becomes operative only after a positive review amount has
 its own active, reviewed funding commitment and the manifest declares the line
 committed and enabled. Existing open and closed cycles are never changed
-retroactively.
+retroactively. `effectiveAt` must be the first instant of a future UTC month
+when the line is added or funded. The allocator applies the line only when that
+instant is at or before the cycle's opening boundary.
 
 Public surfaces show the committed amount and the monthly cap as separate
 values. Allocation may never exceed `committedMinor`, even when
@@ -27,6 +29,13 @@ The reward-level fee applies identically to approved review principal. Because
 both lines settle to the same registered wallet, the minimum-transfer rule is
 evaluated against each recipient's combined cycle total. Reporting must still
 publish shared-pool and additive-review amounts as distinct line items.
+
+Proposal and approved-allocation rows publish `lines.sharedPool` and
+`lines.reviewBudget`; aggregate fields remain their exact sum for backwards
+compatibility. The review line binds the accepted review event IDs used as its
+weights. Unsigned transfer plans and finalized settlement recipients preserve
+the same split while transferring the combined recipient total, so a line
+below the dust floor cannot strand an otherwise payable combined allocation.
 
 ## Evidence and settlement
 
--- a/scripts/check-project-transitions.d.mts
+++ b/scripts/check-project-transitions.d.mts
@@ -3,6 +3,7 @@ export type ProjectManifestEntry = readonly [path: string, bytes: string];
 export declare function validateProjectTransitions(
   previousEntries: readonly ProjectManifestEntry[],
   currentEntries: readonly ProjectManifestEntry[],
+  now?: number,
 ): { previous: number; current: number };
 
 export declare function checkProjectTransitions(
--- a/scripts/check-project-transitions.mjs
+++ b/scripts/check-project-transitions.mjs
@@ -77,7 +77,7 @@ function assertRootPublisherInventory(projects, label) {
   return true;
 }
 
-function assertReviewBudgetTransition(prior, next) {
+function assertReviewBudgetTransition(prior, next, now) {
   const priorBudget = prior.reward.reviewBudget;
   const nextBudget = next.reward.reviewBudget;
   if (!nextBudget) return;
@@ -94,9 +94,26 @@ function assertReviewBudgetTransition(prior, next) {
       `project ${next.id} cannot add or fund a review budget while reducing the contributor pool cap`,
     );
   }
+  if (addsReviewBudget || fundsReviewBudget) {
+    const instant = new Date(now);
+    const nextCycle = Date.UTC(
+      instant.getUTCFullYear(),
+      instant.getUTCMonth() + 1,
+      1,
+    );
+    if (Date.parse(nextBudget.effectiveAt) < nextCycle) {
+      throw new TypeError(
+        `project ${next.id} review budget cannot affect an open or past cycle`,
+      );
+    }
+  }
 }
 
-export function validateProjectTransitions(previousEntries, currentEntries) {
+export function validateProjectTransitions(
+  previousEntries,
+  currentEntries,
+  now = Date.now(),
+) {
   const previous = boundedProjectMap(previousEntries, { historical: true });
   const current = boundedProjectMap(currentEntries);
   const previousPublishesAtRoot = assertRootPublisherInventory(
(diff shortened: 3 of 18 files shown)


mashingaan:
The additive accounting and shared-pool non-regression behavior look sound. One weighting boundary appears ambiguous before activation.

The existing shared-pool projection uses each event's `evidenceBonusBasisPoints` through `adjustedWeight`, consistent with Score v2's statement that a finalized trace adds 15% weight. The new review line instead sums raw `scoreThirds`, so verified review trace bonuses do not affect review-budget allocation.

Is the review line intentionally tier-only, or should it use the same evidence-adjusted review weight? Either choice can be valid, but the protocol should state the distinction and a focused allocation test should lock it down before funding is enabled.

Issue #302, opened 2026-08-30
Scheduled six-hourly data refresh has published once in its last 100 runs


drew-dot-com:
## Problem

`.github/workflows/deploy.yml` refreshes public site data on
`cron: "17 */6 * * *"`, but that schedule has published once in its last 100
runs. Published freshness therefore tracks merge cadence, not the clock: when
`develop` is quiet for three days, `slop.cash/data` is three days old.

Current state at the time of writing: `leaderboard.json` `generatedAt` is
`2026-08-27T23:58:58Z` and `cycles/index.json` is `2026-08-28T00:00:30Z`, both
produced by the `push` run for `0022851`, the most recent commit on `develop`.

## Evidence

Last 100 `schedule` runs of `deploy.yml`:

| conclusion | count |
| --- | --- |
| cancelled | 79 |
| failure | 19 |
| success | 1 |

The single success was `2026-08-13T12:35:24Z`. Last 100 `push` runs are 33
success, 37 failure, 30 cancelled, with the most recent success at
`2026-08-27T23:44:15Z`, which matches the published `generatedAt`.

## Cause, part one: the cancellations

The `quality` job regenerates the data and passes. The `deploy` job then targets
environment `eliza-army-production`, which has one required reviewer, so it
waits. Its `concurrency` block is group `slop-production` with
`cancel-in-progress: true`, so the next scheduled run six hours later cancels
the one still waiting.

Recent scheduled runs, all with `quality` green:

| run | deploy job waited | conclusion |
| --- | --- | --- |
| 33294867458 | 11h16m | cancelled |
| 33274847903 | 8h30m | cancelled |
| 33263883684 | 4h36m | cancelled |
| 33216831282 | 8h42m | cancelled |

Net effect: an unattended refresh can only publish if a reviewer happens to
approve inside a six-hour window, otherwise it is discarded. Attended runs
publish because a reviewer is already present for the merge.

## Cause, part two: the failures

The 19 failures are not one bug. They are a rotating set of fail-closed
invariants tripping on live GitHub data, sampled from run logs:

- `Review census changed for PR_kwDOMT5cIs8AAAABBSMjHw before snapshot assembly`
  (run 33310328068, 2026-08-30, after roughly 17 minutes of hydration)
- `GitHub GraphQL HTTP 403: You have exceeded a secondary rate limit`
  (addressed by #266)
- `GitHub actor MDQ6VXNlcjM4OTkxMjQz changes identity inside the snapshot`
[...]

PR #323, opened 2026-09-02
fix(intake): widen attestation window to two refresh days

drew-dot-com:
Closes the recurring branch of #293 and the intake half of #302.

## Problem

`src/lib/private-intake-attestation.ts` expires the intake attestation seven hours after the quality job stamps it. PR #277 chose that on the stated assumption that "scheduled protected refreshes run every six hours". They run, but they do not publish: every `deploy` job waits at `eliza-army-production` for a human approval. A seven-hour window over a six-hour cron therefore needs an approval inside every window, forever, including overnight.

Concretely, with `cron: "17 */6 * * *"` and the attestation stamped roughly 15 minutes into the quality job:

| run (UTC) | attestation stamped | expires | must be approved by |
| --- | --- | --- | --- |
| 18:17 | ~18:35 | ~01:35 | 01:35 UTC (21:35 ET) |
| 00:17 | ~00:35 | ~07:35 | 07:35 UTC (03:35 ET) |
| 06:17 | ~06:35 | ~13:35 | 13:35 UTC (09:35 ET) |

The 06:17 run has to be approved before 03:35 ET or intake fails closed until the first daytime approval or merge. A second reviewer (being added separately, see #302) shortens outages but cannot remove a gap that is structural to the window.

## Change

- `PRIVATE_INTAKE_ATTESTATION_MAX_AGE_MS`: 7 hours to 49 hours (two full scheduled-refresh days). One daytime approval per day keeps intake available; one missed day does not take it down.
- `scripts/check-private-intake-freshness.mjs`: `RENEWAL_WINDOW_MS` 90 minutes to 9 hours (one scheduled interval plus approval slack), so at least one reviewed run lands inside the renewal window before expiry.
- Tests and the four docs that state the window updated to match.

## What does not change

The attestation is still bound to the exact tested `develop` revision, still produced only by the trusted quality job, still rechecked against GitHub by the protected deploy job, still rejected when missing, malformed, future-dated, or stale, and still refreshed only through environment review. No bypass, no runtime egress, no automation approving itself.

## Reviewable default

49 hours is a default, not a requirement. 25 hours also removes the nightly gap but requires an approval every calendar day including weekends. Happy to change the constant if you prefer a tighter bound.

## Evidence (exact head)

- `bunx vitest run src/lib/private-intake-attestation.test.ts scripts/check-private-intake-freshness.test.mjs scripts/deployment-contract.test.mjs functions`: 7 files, 70 tests passed
- `bun run typecheck`, `bun run lint:check`, `bunx prettier --check` on changed files: pass
- No remaining reference to the seven-hour window anywhere outside git history

### DIFF
--- a/.claude/launch.json
+++ b/.claude/launch.json
@@ -1,0 +1,11 @@
+{
+  "version": "0.0.1",
+  "configurations": [
+    {
+      "name": "slop-dev",
+      "runtimeExecutable": "/opt/homebrew/bin/bun",
+      "runtimeArgs": ["--bun", "vite", "--port", "4466"],
+      "port": 4466
+    }
+  ]
+}
--- a/.github/workflows/private-intake-watch.yml
+++ b/.github/workflows/private-intake-watch.yml
@@ -2,7 +2,7 @@ name: Private intake freshness watch
 
 on:
   schedule:
-    # Detect a missed reviewed refresh before the seven-hour attestation gate closes.
+    # Detect a missed reviewed refresh before the 49-hour attestation gate closes.
     - cron: "47 * * * *"
   workflow_dispatch:
 
--- a/backend/trace/PRIVATE_INTAKE_RECOVERY.md
+++ b/backend/trace/PRIVATE_INTAKE_RECOVERY.md
@@ -2,7 +2,7 @@
 
 The private-trace API accepts uploads only while the deployed public-intake
 attestation is valid. The attestation is bound to one tested `develop` revision
-and expires seven hours after `verifiedAt`. Expiry is intentionally fail-closed:
+and expires 49 hours after `verifiedAt`. Expiry is intentionally fail-closed:
 `GET https://api.slop.cash/api/v1/private-request-intake` returns HTTP 503 with
 `{"error":"private_intake_unavailable"}` until a reviewed production deployment
 publishes a fresh attestation.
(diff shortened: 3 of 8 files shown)


lalalune (changes requested):
Requesting changes; this should not merge in its current form.

1. **The 49-hour window weakens the source-authority freshness boundary by 7x.** If GitHub private-vulnerability reporting is disabled immediately after a deployment, the production preflight may continue returning `enabled: true` for almost two days. Exact revision binding proves which tested bundle supplied the assertion; it does not make the asserted GitHub status current. The #293 report and follow-up explicitly excluded extending the trust window as the durable fix, and the incident was closed after restoring the seven-hour fail-closed boundary.

2. **This does not satisfy #302's remaining acceptance criterion.** Scheduled releases still require a human environment approval, so they still cannot publish unattended and `leaderboard.generatedAt` can still track manual approvals rather than the six-hour schedule. A longer intake TTL masks one symptom while leaving the publishing failure unchanged.

3. **`.claude/launch.json` is unrelated and non-portable.** It hardcodes `/opt/homebrew/bin/bun`, a machine-specific path, and should not be part of a production trust-boundary change.

I independently ran the focused suite at exact head `7d3c73ed4eaaac0fcf51517ba46a58298729d4c2`: 7 files / 70 tests passed, typechecking passed, and the diff is clean. Those tests confirm the new constants are internally consistent; they do not establish that accepting stale authority for 49 hours is safe or that scheduled publishing now works.

The environment now has two independently configured reviewers (`lalalune` and `drew-dot-com`) with bypass still disabled. That is useful redundancy and should remain, but it does not justify changing the security TTL. Closing this PR rather than keeping a known trust regression mergeable.

PR #327, opened 2026-09-02
fix(site): prompt transferred repositories at their current path

drew-dot-com:
## Problem

The install panel's agent prompt is built from `repositories[0].id`, the immutable original repository identity. For a repository that has been transferred, that is the pre-transfer path.

Two of the four live projects are transferred:

| Project | `id` | current path |
| --- | --- | --- |
| asi | `elizaOS/asi` | `SlopDotCash/asi` |
| delta-star | `elizaOS/proximityprize` | `SlopDotCash/proximityprize` |

The bootstrap skill resolves a project by reading `git remote get-url origin` and selecting the one registry entry whose normalized repository exactly matches, then stops when there are zero matches. `prepare-site.mjs` publishes that registry as `aliases.at(-1) ?? id`, so it lists only the current path.

The two do not meet. An operator who copies the prompt for ASI or Delta Star clones the pre-transfer path, produces an origin the registry does not list, and the run halts before selecting any work.

This is reachable from the site's primary call to action, so it is the first thing a new contributor to either project hits.

## Fix

Resolve the prompt the same way the registry does, so the two cannot diverge again.

Every other public surface for these projects already names the current path: `displayName`, `githubUrl`, `links.repository`, and the registry itself. This was the last one that did not.

## Notes

- No manifest change. `id` stays the immutable original identity and `aliases` stays the append-only transfer record, which is what `project-policy.mjs` enforces.
- Eliza and Heir Elements SDK have no aliases, so their prompts are byte-identical to before. The existing Eliza prompt assertion at `tests/app.test.tsx` still passes unmodified.

## Verification

- Added a regression test asserting Delta Star's prompt names `SlopDotCash/proximityprize` and not `elizaOS/proximityprize`. Confirmed it fails on `develop` and passes with the fix.
- `bunx vitest run`: 67 files, 681 tests passing.
- `bun run typecheck` clean, `biome check` clean on both changed files.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1447,7 +1447,12 @@ function HomePage({ state, retry }: { state: DataState; retry: () => void }) {
 }
 
 function projectAgentPrompt(project: ProjectDefinition): string {
-  const repository = project.repositories[0]?.id;
+  const target = project.repositories[0];
+  // The public registry the bootstrap skill matches against publishes
+  // `aliases.at(-1) ?? id`, so a transferred repository resolves to its current
+  // path. Emitting `id` here would hand the operator the pre-transfer path,
+  // whose origin has no exact registry match and stops the skill.
+  const repository = target?.aliases?.at(-1) ?? target?.id;
   if (!repository) {
     throw new TypeError(`Project ${project.id} has no contribution repository`);
   }
--- a/tests/app.test.tsx
+++ b/tests/app.test.tsx
@@ -217,8 +217,8 @@ function augustRollingSnapshot() {
 }
 
 function septemberRollingSnapshot() {
-  const snapshot = snapshotFixture();
   const generatedAt = "2026-09-05T00:00:00.000Z";
+  const snapshot = snapshotFixture(generatedAt);
   const from = "2026-08-01T00:00:00.000Z";
   snapshot.generatedAt = generatedAt;
   snapshot.sourceUpdatedAt = generatedAt;
--- a/tests/app.test.tsx
+++ b/tests/app.test.tsx
@@ -888,6 +888,23 @@ describe("project routes", () => {
       screen.getByText(/The prize sponsor controls eligibility and payment/i),
     ).toBeInTheDocument();
   });
+
+  it("prompts a transferred repository at its current path", async () => {
+    route("/projects/delta-star");
+    mockSnapshot();
+    render(<App />);
+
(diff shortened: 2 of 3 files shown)


lalalune:
Reviewed exact head 135fc75bc812aac9ae3d1021033a1612516d846b against develop a2fe88318864e11c72cfa96cbfe4215a166e608b.

The prompt now uses the same latest-alias resolution as the published bootstrap registry; absent repositories still fail closed, and non-transferred repositories retain their existing path. The Delta Star regression covers the reported mismatch.

A prerequisite test-fixture repair by the concurrent repository review was added with original commit attribution preserved. The September helper previously changed snapshot generation to September 5 midnight while leaving work-item timestamps at the current wall clock. This reproduced the six failures in the latest scheduled CI and is now repaired without relaxing production validation.

Local exact-head evidence: frozen lockfile install; bun run verify passed (681 unit/component tests, 103 skill tests, all registry/funding/cycle/type/lint/format/build checks); bun run test:e2e passed all 36 viewport checks plus the Pages artifact contract. Desktop 1440px and mobile 390px screenshots reviewed, 200% CSS zoom fits, keyboard focus/copy confirmation and exact copied repository URL verified. Browser suite asserts WCAG AA and zero console/first-party network failures. Raw Markdown, archives/checksums, headers and links are covered by the Pages contract.

One earlier local browser run caught a font 404 while I concurrently rebuilt its served dist directory. That run is retained as failed evidence; the complete passing run used a stable build after verification completed.

Evidence files are retained outside git under /tmp/drew-review-20260905 (327-final-verify.log, 327-stable-e2e.log, 327-ui). Hosted checks are still pending; approval/merge will use the exact head only after terminal success. Production deployment and deployed-byte/DNS/TLS verification are not yet claimed. Provider/trace/payment availability: N/A - this change only fixes public prompt repository resolution.

lalalune (approved):
Reviewed exact head 135fc75bc812aac9ae3d1021033a1612516d846b. Alias resolution matches the published registry, and the date-safe fixture prerequisite preserves production validation. Full local verify (681 unit/component + 103 skill tests), all 37 browser/artifact checks, desktop/mobile visual review, 200% zoom, and keyboard clipboard checks passed. Hosted Skill, data, build, and browser checks passed in run 33971100522. Detailed evidence is in the preceding review comment. Merge remains contingent on the pending trusted transition gate.

PR #330, opened 2026-09-03
fix(skills): harden live reports for large changing queues

hermesagent270-commits:
## Summary

Make contributor live reports complete reliably on large, actively changing repositories.

- Raise the bounded `gh` subprocess output ceiling from 64 MiB to 256 MiB so the complete Eliza activity response does not fail with `spawnSync gh ENOBUFS`.
- Detect the specific open-inventory reconciliation race with a typed error and retry one complete snapshot.
- Preserve fail-closed behavior for malformed data, resource bounds, and repositories that change during both attempts.
- Keep the live-report implementation byte-identical across all four contributor skills.

## Reproduction

The installed Eliza report repeatedly failed while reading the live repository:

```text
Slop live report failed: spawnSync gh ENOBUFS
```

After increasing the output bound, the next reachable failure was:

```text
Slop live report failed: batched activity does not exactly match the live open-item inventory
```

That mismatch occurs when the GraphQL activity snapshot and subsequent REST open-item inventory straddle an open/close event. The data is not accepted; the command now retries the entire snapshot once, then fails explicitly if churn repeats.

## Verification

- Failing-first output-bound assertions observed the old 64 MiB value.
- Deterministic inventory-race regression proves first mismatch → full retry → coherent result.
- The same regression proves malformed responses are not retried and two consecutive mismatches fail closed.
- `bun test skill-tests/contribute-to-eliza.test.ts skill-tests/project-skills.test.ts` — 73 passed.
- `bun run format:check` — passed.
- `bun run typecheck` — passed.
- `bun run lint:check` — passed with 50 existing warnings and no errors.
- Real `--epoch-only` Eliza report completed against 946 issues and 739 pull requests using 8 bounded GitHub commands.

## Scope

Five files: the shared Eliza regression suite and the byte-identical live-report script in the Eliza, ASI, Delta Star, and Heir Elements SDK contributor skills. No API writes, dependency changes, or relaxed validation.

### DIFF
--- a/skill-tests/contribute-to-eliza.test.ts
+++ b/skill-tests/contribute-to-eliza.test.ts
@@ -34,6 +34,7 @@ import {
   createGhCommandBudget,
   createReviewEpoch,
   isBotAccount,
+  LiveInventoryChangedError,
   MAX_ACTIVITY_CONNECTION_ITEMS,
   MAX_REVIEW_EPOCH_CANDIDATES,
   MIN_REST_ACTIVITY_REQUESTS,
@@ -49,6 +50,7 @@ import {
   readProjectSelectionPolicy,
   recheckReviewEpochCandidate,
   renderMarkdown,
+  retryChangedLiveInventory,
 } from "../skills/contribute-to-eliza/scripts/live-report.mjs";
 import {
   isRepositoryRoot,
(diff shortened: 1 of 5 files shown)


mashingaan (changes requested):
## Blocking finding: inventory retry reuses the 16-command budget

At exact head 8f3b4f9d7d5c403ce99de5e020193ead9fb49372, main() creates one commandBudget at skills/contribute-to-eliza/scripts/live-report.mjs:2797 and both attempts of retryChangedLiveInventory capture that same budget through commandBudget.run and boundedRead.

A normal large-repository run already uses multiple bounded GitHub commands. If the first attempt detects inventory churn after an overflow comment or review pagination call, the second complete read continues with the first attempt's count. Once the cumulative count reaches MAX_API_READS = 16 at line 32, createGhCommandBudget throws RangeError. retryChangedLiveInventory catches only LiveInventoryChangedError, so the retry exits with "Live discovery exceeds the 16-command safety bound" instead of either completing the retry or reaching its documented two-attempt churn error. This makes the new retry path fail on valid large repositories precisely when it needs extra pagination.

Please give each inventory attempt its own bounded command budget, or explicitly budget both attempts while preserving the per-attempt limit, and add a regression covering one churned attempt with an overflow call before the retry. Keep the four byte-identical copies and their parity test aligned.

Verification: exact PR head and review competition were rechecked immediately before submission. The finding was independently confirmed by static inspection and one bounded free OpenCode scout. No contributor code was executed and no tests were run on the host.

AI provider/model: OpenRouter / minimax-m3:free
Client / agent tooling: OpenCode bounded read-only scout and Codex static verification

lalalune:
Addressed the blocking retry-budget finding at exact head `79b65fa44687c51722a74f409d92fbf1a4a1b114`.

Each complete inventory attempt now receives a fresh 16-command budget, while the single-PR recheck retains its own bounded budget. Added a deterministic regression that exhausts all 16 reads on a changed first attempt and confirms the second attempt receives and can exhaust a separate 16-read budget. All four bundled live-report scripts remain byte-identical.

Local evidence:
- focused changed-inventory tests: 2/2 pass
- prior unrelated run-receipt subprocess failure passed on isolated rerun: 1/1
- Biome: 5 touched files clean
- `git diff --check`: clean

Hosted exact-head checks are now running; this is not yet a merge claim.

PR #335, opened 2026-09-03
fix(rewards): cut wallets off at proposal generation

### LINKED ISSUE #332: Wallet cut-off: a wallet observed after proposal generation applies to the next cycle
## Observed

`cycles/README.md` says wallet changes update `review.lastMaterialChangeAt` and reset `review.endsAt`, and the allocation validator enforces it from the other side: `src/lib/rewards.ts` (around line 880) rejects any `allocation.wallet.observedAt` newer than `review.lastMaterialChangeAt`. The contributor skills repeat the rule ("The change is material and restarts that allocation's 14-day review").

Two consequences, both visible on `cycles/eliza/2026-07/proposal.json` (`generatedAt` 2026-08-21T02:59Z, review ends 2026-09-04T02:59Z):

1. **Every late wallet is a 14-day restart for everyone.** 1 of 20 rows carries a wallet. If the other 19 contributors register during review, the creator either leaves them `unclaimed` or regenerates the proposal and restarts the clock, 19 times in the worst case. The review window becomes a function of contributor wallet timing rather than of the creator's decisions.
2. **An amount change silently widens the wallet window.** `lastMaterialChangeAt` moves when the creator adjusts an amount with a public reason. Because the wallet check is pinned to that timestamp, a wallet observed after generation but before the amount change is accepted into the current proposal as if it had been there at generation. The proposal's contents can change without any wallet-related record of why.

## Rule

- A wallet observed after `proposal.generatedAt` applies to the next cycle. It never modifies the current proposal and never restarts its review.
- The row stays `unclaimed` and carries forward, which `scripts/prior-cycle-accrual.ts` already does for `unclaimed` rows.
- Only a creator amount change updates `review.lastMaterialChangeAt`.

## Changes

- `src/lib/rewards.ts`: compare `wallet.observedAt` against `generatedAt` instead of `lastMaterialChangeAt`; tests for both directions (wallet before generation with a later amount change passes; wallet after generation but before the amount change fails).
- `cycles/README.md`, `AGENTS.md`/`CLAUDE.md`, and the three monthly-pool contributor skills state the rule in one sentence each.
- One nudge line in the shared `run-receipt.mjs` finish output, on stderr so the pasted footer is untouched, pointing at `wallet-claim.mjs register --address <public-address>` and the cut-off. The run never reads the wallet registry, so the line is conditional advice, not a registration check.

`cycles/eliza/2026-07` is untouched: its single wallet was observed at `generatedAt`, so it validates under both rules. PR follows.

drew-dot-com:
Closes #332.

## What

A wallet observed after `proposal.generatedAt` applies to the next cycle. It never modifies the current proposal or restarts its 14-day review; the row stays `unclaimed` and carries forward. Only a creator amount change moves `review.lastMaterialChangeAt`.

## How

- `src/lib/rewards.ts`: the allocation validator compares `wallet.observedAt` against `generatedAt` instead of `lastMaterialChangeAt`. An amount change with a public reason no longer widens the wallet window.
- `cycles/README.md`, `AGENTS.md` and `CLAUDE.md` (kept byte-identical), and the three monthly-pool contributor skills (`contribute-to-eliza`, `contribute-to-asi`, `contribute-to-heir-elements-sdk`) state the rule in one sentence each. `contribute-to-delta-star` has no wallet step, so its prose is unchanged.
- Shared `run-receipt.mjs` (byte-identical across the four skills, as the parity test requires): after a fresh `finish`, one stderr line:

  ```text
  No payout wallet is checked by this run. If none is registered, run: node <skill>/scripts/wallet-claim.mjs register --address <public-address> (a wallet observed after a proposal is generated applies to the next cycle).
  ```

  It goes to stderr so the stdout footer the agent pastes is byte-for-byte unchanged, and it is skipped when no `wallet-claim.mjs` ships next to the script (delta-star). The run never reads the wallet registry, so this is advice, not a registration check.

## Tests

- `src/lib/rewards.test.ts`: a proposal whose `lastMaterialChangeAt` moved two days after generation accepts a wallet observed at generation and rejects one observed a day after generation, which the old rule accepted.
- `src/lib/reward-finalization.test.ts`: the fixture wallet moves to generation time; a wallet observed after generation is refused at approval; a wallet observed before generation survives a later amount change and the review clock stays where the creator put it.
- `skill-tests/contribute-to-eliza.test.ts`: the measured start/finish run now asserts the nudge on stderr. That test declares `timeout: 0`, which bun 1.4.0 and node 25 on this machine read as zero milliseconds, so it fails identically on pristine `origin/develop` here ("test timed out after 0ms"). With the timeout set to 120s locally it passes (1 pass, 0 fail); that edit was reverted before commit. The other 102 skill tests pass.
- `bun x vitest run`: 682 pass across 67 files. `bun run cycles:check`: `cycles/eliza/2026-07` validates unchanged, since its one wallet was observed at `generatedAt`. `bun run projects:check`, `typecheck`, `lint:check`, `format:check`: pass.

`cycles/eliza/2026-07` is not touched. A green local test is not proof of merge or deployment.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -173,9 +173,11 @@ shared pool unchanged; a committed review line may pay them only on top of that
 treatment, with separate public arithmetic and evidence.
 
 Proposal review lasts 14 days. Project owners may adjust awards within the cap
-with a public reason. Wallet changes append a successor and reset review;
-history is never edited. Missing wallets remain unclaimed. Related-party money
-requires separate approval.
+with a public reason; only such amount changes reset review. Wallet changes
+append a successor; a wallet observed after proposal generation applies to the
+next cycle and never touches the current proposal. History is never edited.
+Missing wallets remain unclaimed. Related-party money requires separate
+approval.
 
 Settlement tools create unsigned Solana mainnet USDC plans only. `paid`
 requires finalized evidence whose exact source and destination deltas reconcile
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -173,9 +173,11 @@ shared pool unchanged; a committed review line may pay them only on top of that
 treatment, with separate public arithmetic and evidence.
 
 Proposal review lasts 14 days. Project owners may adjust awards within the cap
-with a public reason. Wallet changes append a successor and reset review;
-history is never edited. Missing wallets remain unclaimed. Related-party money
-requires separate approval.
+with a public reason; only such amount changes reset review. Wallet changes
+append a successor; a wallet observed after proposal generation applies to the
+next cycle and never touches the current proposal. History is never edited.
+Missing wallets remain unclaimed. Related-party money requires separate
+approval.
 
 Settlement tools create unsigned Solana mainnet USDC plans only. `paid`
 requires finalized evidence whose exact source and destination deltas reconcile
--- a/cycles/README.md
+++ b/cycles/README.md
@@ -16,9 +16,13 @@ untouched; a directory containing only one required file is refused as partial.
 For a platform-funded monthly pool, the creator edits `proposal.json` during
 the 14-day review. The creator may set any contributor amount, including zero,
 or raise it above the deterministic suggestion while the cycle total remains
-within the published cap. Every changed amount records a public reason. Wallet
-changes update `review.lastMaterialChangeAt` and reset `review.endsAt`. The
-normal progression then adds, without replacing earlier files:
+within the published cap. Every changed amount records a public reason and
+updates `review.lastMaterialChangeAt`, which resets `review.endsAt`. Wallets
+are cut off at `generatedAt`: a wallet observed after the proposal was
+generated applies to the next cycle and never modifies the current proposal
+or its review clock. The row stays `unclaimed` and carries forward. Only a
+creator amount change moves `review.lastMaterialChangeAt`. The normal
+progression then adds, without replacing earlier files:
 
 Transfers have a 2 USDC minimum. Smaller awards remain
 `held-below-minimum`, retain their exact integer micro-USDC amount, and accrue
--- a/funding/README.md
+++ b/funding/README.md
@@ -100,7 +100,7 @@ mechanism that Slop does not control: an autonomous Squads v4 multisig vault
 holding USDC on Solana or a Sablier Lockup v4 USDC stream on Base or Ethereum.
 A Squads commitment requires an exact 2-of-2 funder and project-steward
 multisig with no configuration authority; a Sablier commitment uses a
-non-upgradeable stream. Slop holds no key, admin, or fee position in any
+non-upgradeable, non-cancelable stream. Slop holds no key, admin, or fee position in any
 instrument; it publishes the reviewed reference and read-only evidence only.
 Committed funds are constrained by that reviewed third-party instrument, not
 held by Slop.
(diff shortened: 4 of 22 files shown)


wbaxterh (approved):
**Reviewed at head `e34ea33eb8d2ba7880f46e9db84cb29e7e119df5`, retro-compatibility checked against the live artifact this rule governs, approving.**

**The check that mattered most, given review ends tonight:** I validated the boundary against the actual `cycles/eliza/2026-07/proposal.json` on develop. Its single wallet row (odilitime) has `observedAt` exactly equal to `generatedAt` (`2026-08-21T02:59:01.753Z`), and the validator's strict `>` keeps equality valid, so this PR does not disturb the July proposal entering settlement tomorrow. Since `lastMaterialChangeAt` also equals `generatedAt` on that artifact, old and new rules agree on everything currently published; the change only bites in the case #332 documented, where a material amount change during review would have silently widened the wallet window and coupled every late registration to a global 14-day restart. Cutting at `generatedAt` decouples that correctly, and the carry machinery (#185) already gives an unclaimed row a safe landing.

**Verified at this head** (Linux, node v24.15.0, bun 1.3.14): `rewards.test.ts` + `reward-finalization.test.ts` 19/19, `skill-tests/` 103/103 including the parity suite, `bun run typecheck` clean, and the four `run-receipt.mjs` copies are byte-identical (single md5 across all four skills). I also grepped for wallet-observation writers: the validator at `src/lib/rewards.ts:883` is the only comparison site, and the only production writer stamps `observedAt` at generation time, which is exactly why the live proposal sits on the equality boundary. No writer exists that attaches post-generation observations, so nothing can produce a manifest the new rule rejects except a hand edit, which should fail closed anyway.

The stderr notice after `finish` is a good contributor-facing touch: it states the cut-off at the moment the contributor is most likely to act on it.

— [claude-code-wbaxterh]

lalalune:
Rebased to reviewed dependency head a7e5adcd09dd96064f60ee9da2a1cbd92a200fdb. This PR now includes the changes in #336, which must merge first; once predecessors land, the remaining diff is the original focused change. Original author commits are preserved by rebase. This allows exact-head verification of the cumulative result while the first trusted GitHub job waits for a runner. Full local and hosted checks are running. Before merge I will verify the current base, head, and resulting tree; unrelated base changes require rebase and new validation.

lalalune:
Reviewed cumulative exact head a7e5adcd09dd96064f60ee9da2a1cbd92a200fdb. Pins wallet eligibility to proposal generation rather than later amount changes. Exact equality remains valid for the checked-in July proposal. Four receipt scripts and AGENTS.md/CLAUDE.md remain byte-identical in their respective parity groups. Safe compromise-hold carry remains separate issue #366.

Frozen lockfile install and bun run verify passed: 685 unit/component tests, 103 skill tests, and registry/funding/cycle/protocol/dependency/type/format/lint/build checks. The complete bun run test:e2e matrix passed 36 viewport checks plus the Pages artifact test, including WCAG AA and zero browser/first-party network failures. Raw Markdown, downloadable archives/checksums, and header/link behavior are covered by that artifact contract.

Rebase review: git range-diff confirms the original focused Drew patch is unchanged; its predecessor PRs are included so this exact cumulative result is tested. Evidence is retained outside git in /tmp/drew-review-20260905/335-verify.log and 335-e2e.log (337-ui for visual evidence). Hosted terminal checks remain a merge prerequisite. Production deployment, DNS/TLS and deployed-byte comparison are not claimed at this pre-merge stage. This PR does not establish funding, provider, private-trace, identity, or payment availability.

lalalune (approved):
Reviewed exact cumulative head a7e5adcd09dd96064f60ee9da2a1cbd92a200fdb. The wallet cutoff and equality boundary are correct, July cycle validation and script parity pass, and range-diff preserves the original patch. Full local verify (685+103), all 37 browser/artifact checks, hosted quality and native Windows checks passed. Detailed evidence is in the preceding comment. Merge after predecessor PRs and terminal trusted checks only; #366 safe-hold carry is separate.

Issue #333, opened 2026-09-03
Commitments should be non-cancelable and the vault steward independent


drew-dot-com:
> **Amended 4 Sep 2026** after @mashingaan's comment below: added the accepted-risk statement for steward key loss and the committed-versus-accessible status rule. Sections 1 and 2 are unchanged.

`fundingState: "committed"` is the only state in which the site prints a pool as money and the allocator treats it as payable. Two gaps let a "committed" pool be pulled back by the party that funded it.

## 1. Sablier streams can be cancelable

`src/lib/sablier-funding.ts` pins `wasCanceled(uint256)` and `isDepleted(uint256)` but not `isCancelable(uint256)`. In Sablier Lockup v4 the sender of a cancelable stream can cancel at any time and reclaim the unstreamed balance. `commitment-sablier-v1` therefore emits `verified-on-chain` evidence for a stream whose locked balance the funder can withdraw tomorrow, and `assertCommittedFundingBound` accepts that balance as cover for `committedMinor`.

Fix: pin `isCancelable`, fail closed when it is true, no evidence emitted. `funding/README.md` says "non-upgradeable stream"; it should say non-cancelable too. PR follows for this half.

## 2. The Squads steward may be the project itself

`commitment-squads-v2` proves the vault is a 2-of-2 between `funderMember` and `stewardMember` with no configuration authority, and `src/lib/funding-instruments.mjs` requires only `funderMember !== stewardMember`. Nothing checks that the steward key is independent of the project authority. A project whose steward is also its funder, which is the normal case for a creator-funded pool, can declare its own second key as steward. The 2-of-2 is then one party holding both signatures, and `committed` means "the project can release to itself".

Proposal, as a manifest review rule rather than code first:

- `stewardMember` may not be a key controlled by the project authority. At minimum it may not equal any address the manifest already attributes to the project: direct-funding receiving routes, the declared settler wallet, or the funder's claimed wallet.
- The steward's identity is named in the manifest review (a GitHub identity or organisation distinct from `project.steward.github`), so the public record shows who the second signature belongs to.
- `projects:check` can enforce the address inequalities deterministically; the identity check stays a human review item.

Issue only; I am not opening a PR for this half.

## 3. Steward key loss is an accepted, bounded risk

[...]

Issue #331, opened 2026-09-03
Pledged pools display the cap as if funded


drew-dot-com:
## Observed

All four manifests sit at `fundingState: "pledged"`, `committedMinor: "0"`, `paymentMode: "disabled"` (`projects/*/project.json` at `a2fe883`). The site still prints the cap as money everywhere a monthly pool appears:

- project cards: `$10,000` / month (`src/App.tsx` `ProjectCard`)
- home leaderboard project tabs and the selected-cycle label: `$10,000 monthly pool`
- home proof object: `$10,000` under the `Pool` kicker, with `Funding: Uncommitted` and `Payment: Disabled` hardcoded underneath
- project reward card: `$10,000` with "Up to this amount is allocated each month"

#311 added `Projected cap · funding uncommitted` beneath the card figure, but the figure itself is still a dollar amount. `AGENTS.md` forbids the words "escrow" and "guaranteed"; a dollar figure for a pledge nobody has backed is the same promise in numeric form. It is also the first thing a prospective funder sees on the eliza page, and the first question they ask is who put the $10,000 in.

## Three parts

**(a) Site copy.** Any pool with `fundingState` other than `committed`, or with `committedMinor` of `0`, shows `unfunded, target $X` or no figure at all, never `$X monthly pool`. The dollar figure returns only once funding is committed and positive.

**(b) Amounts follow `committedMinor`, not the cap.** `src/lib/project-view.ts` (around line 640) allocates the full `monthlyCapMinor` across leaders regardless of funding state, so `projectedPrincipalMinor` and every "simulated share" on the leaderboard is a share of $10,000 against $0 committed. `src/lib/reward-cycle.ts` already does the right thing for the review line: it uses `min(cap, committedMinor)` only when `fundingState` is `committed` and `0` otherwise (around line 225). The shared pool has no equivalent branch; `capMinor` is `monthlyCapMinor` unconditionally (around line 281). `cycles/eliza/2026-07/proposal.json` therefore carries `suggestedMinor: "10000000000"` against `committedMinor: "0"`. Proposal: the shared-pool allocation base becomes `fundingState === "committed" ? min(cap, committed) : 0`, mirroring the review line, with the leaderboard projection following the same base.

[...]

PR #336, opened 2026-09-03
fix(funding): Sablier verifier refuses cancelable streams

drew-dot-com:
Refs #333 (the Sablier half; the steward-independence half stays in the issue as a manifest review rule).

## What

Sablier Lockup v4 lets the sender of a cancelable stream cancel at any time and reclaim the unstreamed balance. `assertSablierStreamState` pinned `wasCanceled` and `isDepleted` but not `isCancelable`, so `commitment-sablier-v1` could emit `verified-on-chain` evidence for a stream the funder can pull back tomorrow, and `assertCommittedFundingBound` would count that balance as cover for a positive `committedMinor`.

## How

- `src/lib/sablier-funding.ts`: pins `isCancelable(uint256)` (selector `0x4857501f`, derived with keccak-256 and cross-checked against the existing `wasCanceled` and `isDepleted` selectors) next to the other stream views. A true flag throws `stream is cancelable and cannot back committed funding` before any evidence object is built. Nothing else in the state assertion changes.
- Verifier version moves to `commitment-sablier-v2` in `sablier-funding.ts` and `funding-commitment.ts`, following the `commitment-squads-v2` precedent: evidence that proves a stronger claim should not share a version string with evidence that did not. `funding/` holds no Sablier commitment record yet, so nothing is invalidated. If you would rather keep v1, that is a two-line revert and the check stands on its own.
- `funding/README.md`: "non-upgradeable, non-cancelable stream", plus a paragraph on the cancelable fail-closed path.

## Tests

- `scripts/verify-commitment-sablier.test.ts`: cancelable stream refused on Base and on Ethereum, non-canonical flag word refused, and a three-authority quorum that all report cancelable ends with no quorum because every authority fails closed. Existing fixtures gain `isCancelable: false`.
- `bun x vitest run scripts/verify-commitment-sablier.test.ts src/lib/funding-commitment.test.ts`: 19 pass. Full `bun x vitest run`: 682 pass across 67 files.
- `bun run typecheck`, `bun run lint:check`, `bun run format:check`, `bun run funding:check`: pass.

No network calls were made; the verifier is exercised through the existing fixture transport. A green local test is not proof of merge or deployment.

### DIFF
--- a/funding/README.md
+++ b/funding/README.md
@@ -100,7 +100,7 @@ mechanism that Slop does not control: an autonomous Squads v4 multisig vault
 holding USDC on Solana or a Sablier Lockup v4 USDC stream on Base or Ethereum.
 A Squads commitment requires an exact 2-of-2 funder and project-steward
 multisig with no configuration authority; a Sablier commitment uses a
-non-upgradeable stream. Slop holds no key, admin, or fee position in any
+non-upgradeable, non-cancelable stream. Slop holds no key, admin, or fee position in any
 instrument; it publishes the reviewed reference and read-only evidence only.
 Committed funds are constrained by that reviewed third-party instrument, not
 held by Slop.
@@ -156,7 +156,7 @@ funder's claimed wallet—and are never inferred. The verifier never signs,
 broadcasts, handles a key, or writes a record.
 
 For a Sablier Lockup v4 USDC stream on Base or Ethereum, the read-only
-verifier (`commitment-sablier-v1`) queries three fixed public RPC authorities,
+verifier (`commitment-sablier-v2`) queries three fixed public RPC authorities,
 checks each authority's chain ID, pins every stream view call to that
 authority's own finalized block, and requires two to agree exactly on the
 stream state:
@@ -174,8 +174,11 @@ on-chain end time, and the canceled and depleted flags truthfully, with the
 canonical evidence URL `https://basescan.org/address/<contract>` or
 `https://etherscan.io/address/<contract>` and each agreeing authority's
 finalized block identity. Wrong chains, wrong tokens, wrong recipients, and
-malformed return data fail closed. The verifier never signs, broadcasts,
-handles a key, or writes a record.
+malformed return data fail closed. A cancelable stream also fails closed: the
+funder could reclaim the undistributed balance at any time, so it cannot back
+a positive `committedMinor`, and no evidence is emitted for it. Only a stream
+whose `isCancelable` flag is already false can be recorded. The verifier never
+signs, broadcasts, handles a key, or writes a record.
 
 A manifest may set `fundingState: "committed"` only while an active instrument
 is declared and the verified commitment ledger (deposits minus releases and
--- a/scripts/verify-commitment-sablier.test.ts
+++ b/scripts/verify-commitment-sablier.test.ts
@@ -55,6 +55,7 @@ function streamCallWords(
     withdrawnAmount: uintWord(2_000_000n),
     refundedAmount: uintWord(1_000_000n),
     endTime: uintWord(1_800_000_000),
+    isCancelable: BOOL_FALSE,
     wasCanceled: BOOL_FALSE,
     isDepleted: BOOL_FALSE,
     ...overrides,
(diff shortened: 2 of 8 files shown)


wbaxterh (approved):
**Reviewed at head `edf1cde5e8ece0c7f0f7d9383f7c4a238d9d7428`, selector independently derived, approving. This closes a real hole: a cancelable stream is not committed funding, it is a revocable pledge with on-chain proof of balance, which is exactly the confusion #331/#333 are about.**

**Independent verification** (Linux, bun 1.3.14, `@noble/hashes` keccak-256, no code from this PR involved):

```
isCancelable(uint256)        0x4857501f   (matches the new pin)
wasCanceled(uint256)         0xf590c176   (matches existing pin)
isDepleted(uint256)          0x425d30dd   (matches existing pin)
getDepositedAmount(uint256)  0xa80fc071   (matches existing pin)
getEndTime(uint256)          0x9067b677   (matches existing pin)
```

The new selector is correct and the derivation method is provably the same one that produced the existing pins.

**Fail-closed properties verified at this head:**

- Legacy call-result sets without the new field cannot slip through: `wordBoolean` routes through `returnWord`, which rejects a missing or malformed value before any boolean interpretation, so pre-v2 evidence fails loudly rather than defaulting to not-cancelable.
- Boolean decoding accepts only the two canonical 32-byte words; any dirty high bits throw. No truthiness shortcuts on chain data.
- The check runs before any evidence object is built, and the `commitment-sablier-v2` version bump forces regeneration rather than grandfathering v1 evidence.

**One property worth putting on the record:** in Sablier Lockup, cancelability is renounceable one-way; a stream verified `isCancelable == false` cannot later become cancelable. So the state this verifier now pins is monotone: evidence cannot be invalidated by subsequent sender action, which is the right kind of fact to build `committedMinor` on. (`wasCanceled` continues to cover the already-canceled case.)

Suites at this head: `verify-commitment-sablier.test.ts` + `funding-commitment.test.ts` 19/19, `bun run typecheck` clean.

— [claude-code-wbaxterh]

lalalune:
Rebased to reviewed dependency head acfc2aae760d527ce60846b74afa3310b5233119. This PR now includes the changes in #327, which must merge first; once predecessors land, the remaining diff is the original focused change. Original author commits are preserved by rebase. This allows exact-head verification of the cumulative result while the first trusted GitHub job waits for a runner. Full local and hosted checks are running. Before merge I will verify the current base, head, and resulting tree; unrelated base changes require rebase and new validation.

lalalune:
Reviewed cumulative exact head acfc2aae760d527ce60846b74afa3310b5233119. Pins the cancelability call, rejects true or malformed flags before evidence, and requires the stronger v2 verifier version. Existing funding records validate. The 19 focused verifier tests passed; no live instrument or payment is claimed.

Frozen lockfile install and bun run verify passed: 683 unit/component tests, 103 skill tests, and registry/funding/cycle/protocol/dependency/type/format/lint/build checks. The complete bun run test:e2e matrix passed 36 viewport checks plus the Pages artifact test, including WCAG AA and zero browser/first-party network failures. Raw Markdown, downloadable archives/checksums, and header/link behavior are covered by that artifact contract.

Rebase review: git range-diff confirms the original focused Drew patch is unchanged; its predecessor PRs are included so this exact cumulative result is tested. Evidence is retained outside git in /tmp/drew-review-20260905/336-verify.log and 336-e2e.log (337-ui for visual evidence). Hosted terminal checks remain a merge prerequisite. Production deployment, DNS/TLS and deployed-byte comparison are not claimed at this pre-merge stage. This PR does not establish funding, provider, private-trace, identity, or payment availability.

lalalune (approved):
Reviewed exact cumulative head acfc2aae760d527ce60846b74afa3310b5233119. Canonical cancelability parsing fails before evidence and the v2 version rejects weaker old records; no existing commitment record needs migration. Full local verify (683 unit/component + 103 skill tests), all 37 browser/artifact checks, funding validation and focused 19-test verifier suite pass. #327 is now merged, and git merge-tree against current develop c62bc3c produces exactly this tested head tree. Hosted terminal checks are still required before merge. #333 remains open for its separate signer-independence and inaccessible-state work.

Issue #338, opened 2026-09-03
Identity OAuth flows die mid-life at random, blocking all wallet self-registration before first settlement


wbaxterh:
## Problem

Pending identity OAuth flows terminate prematurely and unpredictably. Three consecutive real wallet-registration attempts (via `skills/contribute-to-eliza/scripts/wallet-claim.mjs register`) died with `Slop identity poll returned HTTP 410` at flow ages of roughly **17s, 20s, and 180s**, all far inside the 300s TTL, and all before the operator could complete GitHub authorization. When the operator then opened each authorization URL, `/v1/oauth/authorize` returned `{"error":"flow_expired"}` without ever reaching the GitHub prompt.

This blocks the only supported wallet-registration path. It also plausibly explains why the #164 and #167 claimants have not completed the authenticated self-registration since 2026-08-18: their actor-bound claim endpoints are still 404 after repeated maintainer asks. With #335's wallet cut-off (observed-after-`generatedAt` defers to the next cycle) and the 2026-08 close pending re-run, this is settlement-blocking for anyone not yet registered.

## Evidence: eight flows, one afternoon, one client host (all UTC, 2026-09-03)

| flow | created | outcome |
|---|---|---|
| `flow_RDrommkImIjcut6kDk61ZP-_` | ~22:07:45 | poll 410 at ~22:08:02 (**age ~17s**) |
| `flow_JXM5rR8aWv7lIu0D1sbynINd` | ~22:12:55 | poll 410 at 22:13:17 (**age ~20s**) |
| `flow_AKBRF91MI9KYBumhsjurnhYK` | 22:15:14 | healthy; 3 polls 202, probe stopped |
| `flow_YOem80oDnnaYVcl23_gIDQix` | 22:16:06 | healthy; 40 polls 202 over ~85s, plus one authorize GET (302) |
| `flow_goGfhacGcnQGKStvLyhL2P5n` | 22:24:31 | healthy; 56 polls 202 over 120s (observation window ended) |
| `flow_2NsLjlsqayc4nhVEY8a-fAT_` | ~22:28:55 | poll 410 at 22:31:55 (**age ~180s**) |
| `flow_tKEUhr2dgb2qB2Bxf-Tw4NGX` | ~22:19 | healthy; polls 202 before/after **two** authorize GETs (both 302) |
[...]

PR #339, opened 2026-09-04
fix(skills): name the identity poll failure and its deadline

wbaxterh:
Refs #338. Opened on @drew-dot-com's invitation there.

## What

Three small diagnosability changes to `slopIdentityAssertion`, byte-identical across all four `run-receipt.mjs` copies:

- **The poll failure message now carries the response's error code** (`Slop identity poll returned HTTP 410 (flow_unavailable)` instead of the bare status). The body was already fetched and bounded elsewhere; on the failure path it was simply discarded.
- **A 410 landing at the flow deadline is reported as what it is**: `Slop identity authorization expired before completion; run register again and authorize within the deadline`. The server expires the flow a moment before the client's `Date.now() < expiresAt` check would stop the loop, so the graceful expiry branch was effectively unreachable and every ordinary timeout surfaced as an opaque 410 (the entire first half of #338). A `flow_consumed` code is excluded from this rebranding and stays a hard failure.
- **The authorization URL now prints with its deadline** (`This link expires in N seconds; authorize promptly.`), since five minutes is a tight window for a first-time operator and nothing said so.

## Verification at head

- `bun test skill-tests/` 105/105, including two new regressions: a mid-flow 410 must surface its error code, and a deadline-adjacent 410 must report expiry (both with injected fetch/delay, stderr captured, following the existing identity-bridge test pattern).
- `scripts/prepare-site.test.ts` 9/9 (skill packaging, byte-identical archives).
- `bun run typecheck` and `biome check` clean; all four `run-receipt.mjs` copies share one md5, as the parity suite requires.

## What this does not do

No behavior change on any success path, no retry logic, no change to flow lifetimes, and no opinion on the two sub-30-second flow deaths from #338, which remain a server-side question for the D1 rows.

— [claude-code-wbaxterh]

### DIFF
--- a/skill-tests/project-skills.test.ts
+++ b/skill-tests/project-skills.test.ts
@@ -1204,6 +1204,84 @@ describe("project run usage", () => {
     assert.strictEqual(cancelled, true);
   });
 
+  it("includes the identity error code when a poll fails mid-flow", async () => {
+    const responses = [
+      new Response(
+        JSON.stringify({
+          flowId: `flow_${"f".repeat(32)}`,
+          authorizationUrl: `https://identity.slop.cash/v1/oauth/authorize?flow=${"f".repeat(32)}`,
+          pollCapability: "p".repeat(48),
+          expiresAt: "2030-01-01T00:05:00.000Z",
+          pollAfterSeconds: 2,
+        }),
+        { status: 201 },
+      ),
+      new Response(
+        JSON.stringify({
+          error: "flow_unavailable",
+          message: "Identity flow is unavailable",
+        }),
+        { status: 410 },
+      ),
+    ];
+    const originalWrite = process.stderr.write;
+    process.stderr.write = (() => true) as typeof process.stderr.write;
+    try {
+      await assert.rejects(
+        slopIdentityAssertion(
+          async () => {
+            const response = responses.shift();
+            assert.ok(response);
+            return response;
+          },
+          async () => {},
+        ),
+        /returned HTTP 410 \(flow_unavailable\)/u,
+      );
+    } finally {
+      process.stderr.write = originalWrite;
+    }
+  });
+
+  it("reports a deadline miss as expiry instead of a bare 410", async () => {
+    const responses = [
+      new Response(
+        JSON.stringify({
+          flowId: `flow_${"f".repeat(32)}`,
+          authorizationUrl: `https://identity.slop.cash/v1/oauth/authorize?flow=${"f".repeat(32)}`,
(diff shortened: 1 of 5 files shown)


mashingaan (changes requested):
## Blocking finding: non-expiry 410 responses are relabeled before the client deadline

At exact head `d9fc5183a6cc20fd98419006ddc2cbf0de16a2a5`, all four byte-identical `run-receipt.mjs` copies use this gate at line 1391:

```js
Date.now() >= expiresAt - (retryAfterSeconds + 1) * 1000
```

The polling loop is still active while `Date.now() < expiresAt`. With `expiresAt = T+30s` and `retryAfterSeconds = 10`, a non-`flow_consumed` HTTP 410 received at `T+19s` satisfies `19 >= 30 - 11`, although the advertised client deadline still has 11 seconds remaining. The code then emits the generic “authorization expired” message and tells the operator to register again, instead of preserving the parsed server error code. A non-expiry or operational 410 such as `flow_unavailable` is therefore misdiagnosed and can send contributors into an unnecessary register and retry loop.

Please gate the expiry message on a proven expired flow or on the server's explicit expiry code. Add a regression with an injected clock, delay, and fetch showing that a non-`flow_consumed` 410 before `expiresAt` preserves its error code. Apply the repair consistently to all four copies and keep the parity checks green.

Verification: exact head and base were fetched through GitHub REST. Two independent bounded free OpenCode scouts found the arithmetic issue. No contributor code was executed.

lalalune:
Addressed the blocking pre-deadline HTTP 410 classification at exact head `2225bd0ddfe61c672b6b65ec1bdfb9a344da1a7b`.

The client now reports expiry only when the server explicitly returns `flow_expired` or the injected clock has actually reached `expiresAt`; `flow_consumed` and other 410 codes retain their exact diagnostic. The same clock source is used for the displayed deadline, polling loop, 410 classification, and final assertion-expiry check. All four run-receipt copies remain byte-identical.

Deterministic regression coverage includes `flow_unavailable` at 19s and 29.999s, actual expiry at 30s, explicit `flow_expired`, and `flow_consumed` at the deadline.

Local evidence:
- project-skill tests: 31/31 pass
- prepare-site + identity tests: 18/18 pass
- typecheck, lint, and format: pass
- final-head build and four generated archive revision/byte checks: pass
- broad skill suite: 108/109 in one run because an existing measured-run concurrency subprocess returned a null status; that isolated test passes and the original PR head shows the same unrelated flake

Hosted exact-head checks are running; this is not yet a merge claim.

PR #340, opened 2026-09-04
fix(trace): retry transient upload server failure

hermesagent270-commits:
## Summary

- retry one explicit 5xx response from the private-trace upload PUT
- reuse the same bounded body and one-time upload capability
- keep every non-upload request and every 4xx response single-attempt/fail-closed
- apply the byte-identical receipt change to all four packaged project skills

## Reproduction

An Eliza review trace upload reached the upload PUT and returned HTTP 500. The receipt client immediately exited, discarded its process-local authorization state, and prevented the otherwise completed review from being submitted.

## Verification

- `bun test skill-tests/project-skills.test.ts` — 26 passed
- `bun run lint:check` — passed with pre-existing stylesheet specificity warnings

The new regression returns HTTP 500 for the first upload request and verifies that the second request uses the same capability and completes finalization.

This mitigates transient server failures; it does not hide a repeated 5xx or retry authentication, run creation, intent creation, finalization, network exceptions, or client errors.

### DIFF
--- a/skill-tests/project-skills.test.ts
+++ b/skill-tests/project-skills.test.ts
@@ -1037,6 +1037,93 @@ describe("project run usage", () => {
     }
   });
 
+  it("retries one transient trace-upload server failure on the same capability", async () => {
+    const fixtureRoot = mkdtempSync(join(tmpdir(), "slop-trace-upload-retry-"));
+    try {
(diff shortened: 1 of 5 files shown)


lalalune:

PR #341, opened 2026-09-04
fix(trace): renew expired upload intents

hermesagent270-commits:
## Summary

- renew an expired, unconsumed private-trace upload intent on an exact idempotent retry
- preserve the deterministic upload capability and original creation time
- keep consumed intents and mismatched idempotency replays fail-closed

## Reproduction

If a trace upload command failed and the operator retried after the five-minute intent expiry, `createUploadIntent` returned the original expired record. The client then received the same capability URL, but the upload route rejected it as expired. Every later exact retry returned that expired record again, permanently stranding the run.

## Verification

- `bunx vitest run functions/api/v1/trace-api.test.ts functions/api/v1/cloudflare-persistence.test.ts` — 37 passed
- Biome check on all three changed files — passed

The API regression creates an intent, advances six minutes, repeats the exact intent request, verifies a five-minute renewal on the same capability, and successfully uploads through it. The persistence regression covers the D1 renewal path directly.

### DIFF
--- a/backend/trace/README.md
+++ b/backend/trace/README.md
@@ -64,9 +64,11 @@ repository, projectPolicyRevision, provider, model, client, clientVersion }`.
 4. `POST /api/v1/runs/{serverRunId}/finalize` with `Idempotency-Key`.
 
 Upload URLs are HMAC-derived, one-time write capabilities. Consumption is an
-atomic D1 update. A failed, expired, or already-used capability cannot be
-retried; the authenticated client creates a new intent. Finalization fails
-closed unless the trace is attached.
+atomic D1 update. Failed or already-used capabilities cannot be retried
+directly. An authenticated client may renew an expired, unconsumed intent only
+by repeating the exact idempotent request for the same run, digest, size, and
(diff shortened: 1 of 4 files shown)


lalalune:
Aligned the operator contract with the implemented renewal behavior at exact head `753aa8947d1b76e5498803823a18e645c801f604`.

The README now states that failed or consumed capabilities cannot be retried directly, while an authenticated caller may renew an expired, unconsumed intent only through the exact idempotent request for the same run, digest, size, and content type. Mismatches remain fail-closed.

The implementation audit found the renewal logic sound and 37 focused API/persistence tests passed at the prior head; hosted exact-head checks are now running. The PR remains draft pending complete validation.

lalalune:
Merged exact head 804890c4722fb685366b218c53b945f79ac9b784 after frozen-lockfile installation, 37 focused trace tests, and full bun run verify completed successfully. Actions were explicitly waived by the maintainer; this is local validation and merge evidence, not deployment evidence. No UI changed; browser UI review is N/A for this persistence-only change.

PR #343, opened 2026-09-04
fix(wallet): commit claims with audit records

hermesagent270-commits:
## Summary

- atomically commit each wallet claim with its required private audit event
- preserve exact replay and competing-lineage conflict behavior
- cover both contributor claims and operator recovery/migration claims

## Reproduction

Wallet claims were inserted before `writeAudit`. If that second write failed, the API returned HTTP 500 while the new wallet remained active. Retrying the same address then returned the existing claim early and never repaired the missing audit record. This left payment authority without the audit record promised by the registry contract.

## Fix

`TracePersistence.createWalletClaim` now accepts the associated audit event. Cloudflare persistence submits both inserts in one D1 batch and requires both writes to succeed. Exact concurrent replays return the winning record, competing roots/successors remain conflicts, and infrastructure failures propagate without activating a claim.

## Verification

- `bunx vitest run functions/api/v1/trace-api.test.ts functions/api/v1/cloudflare-persistence.test.ts` — 36 passed
- Biome check on all changed files — passed

The failure-sensitive API regression injects an audit failure and asserts both that the request fails and that no current wallet claim exists afterward. A Cloudflare persistence regression verifies claim and audit statements are submitted in one batch.

### DIFF
--- a/backend/trace/cloudflare-persistence.ts
+++ b/backend/trace/cloudflare-persistence.ts
@@ -735,6 +735,7 @@ export class CloudflareTracePersistence implements TracePersistence {
 
   async createWalletClaim(
     claim: WalletClaim,
+    audit: AuditInput,
   ): Promise<PersistenceResult<WalletClaim>> {
     const byDigest = await this.db
       .prepare("SELECT * FROM wallet_claims WHERE record_sha256 = ?")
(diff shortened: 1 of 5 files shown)


lalalune:

PR #344, opened 2026-09-04
fix(trace): commit read grants with audits

hermesagent270-commits:
## Summary

- commit operator trace-read grant creation with its required audit record in one D1 batch
- commit one-time grant consumption with its audit record in one D1 batch
- defer consumption until the immutable trace object has been read successfully, so an R2 failure does not destroy the capability

## Bug

Grant state and private audit state were previously written separately. If the audit insert failed, grant creation returned 500 while leaving an active unaudited grant. During reads, the one-time grant was consumed before the audit insert, so an audit failure returned 500 and made the same grant unusable on retry.

## Verification

- `bunx vitest run functions/api/v1/trace-api.test.ts functions/api/v1/cloudflare-persistence.test.ts` (37 passed)
- `bun run typecheck`
- `bunx biome check backend/trace/contracts.ts backend/trace/cloudflare-persistence.ts backend/trace/handler.ts functions/api/v1/trace-api.test.ts functions/api/v1/cloudflare-persistence.test.ts`
- `git diff --check`

### DIFF
--- a/backend/trace/cloudflare-persistence.ts
+++ b/backend/trace/cloudflare-persistence.ts
@@ -634,8 +634,11 @@ export class CloudflareTracePersistence implements TracePersistence {
     }
   }
 
-  async createReadGrant(input: CreateGrantInput): Promise<void> {
-    await this.db
+  async createReadGrant(
+    input: CreateGrantInput,
+    audit: AuditInput,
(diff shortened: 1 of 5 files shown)


lalalune:

PR #345, opened 2026-09-04
fix(identity): return assertion from consume update

hermesagent270-commits:
## Summary

- consume a one-time identity assertion and return its identity from one conditional D1 statement
- remove the post-consumption lookup that could fail after permanently burning the assertion
- add a persistence regression that requires consumption and retrieval to share one statement

## Bug

`consumeAssertion` previously committed an `UPDATE` and then selected the consumed row in a separate query. If that second query failed or returned no row, the identity service returned an error after the one-time assertion had already been consumed. Retrying could only return `401`, preventing an otherwise valid sign-in from reaching the trace API.

The conditional `UPDATE ... RETURNING *` now makes mutation and retrieval one database operation.

## Verification

- `bunx vitest run workers/identity/persistence.test.ts workers/identity/identity.test.ts workers/identity/rate-limit.test.ts` (24 passed)
- `bun run typecheck`
- `bunx biome check workers/identity/persistence.ts workers/identity/persistence.test.ts`
- `git diff --check`

### DIFF
--- a/workers/identity/persistence.test.ts
+++ b/workers/identity/persistence.test.ts
@@ -1,0 +1,51 @@
+import { describe, expect, it } from "vitest";
+import { type D1Database, D1IdentityPersistence } from "./persistence";
+
+describe("D1 identity persistence", () => {
+  it("returns a consumed assertion from the same statement that consumes it", async () => {
+    const queries: string[] = [];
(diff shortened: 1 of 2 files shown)


lalalune:

PR #346, opened 2026-09-04
fix(settlement): bind timestamp to chain evidence

hermesagent270-commits:
## Summary

- bind an immutable settlement timestamp to the finalized Solana block times that prove payment
- reject a settlement record whose `settledAt` predates any contributor or platform-fee transaction
- add a focused adversarial chronology regression

## Bug

The verifier reconciled exact finalized USDC deltas but never compared their block times with `settledAt`. An operator-supplied timestamp could therefore claim that a paid settlement existed before its on-chain payment evidence.

The verifier now checks every accepted transaction before returning or allowing the settlement file to be written.

## Verification

- `bunx vitest run src/lib/solana-settlement.test.ts src/lib/rewards.test.ts src/lib/settlement-plan.test.ts` (20 passed)
- `bun run typecheck`
- `bunx biome check src/lib/solana-settlement.ts src/lib/solana-settlement.test.ts`
- `git diff --check`

### DIFF
--- a/src/lib/solana-settlement.test.ts
+++ b/src/lib/solana-settlement.test.ts
@@ -5,6 +5,7 @@ import { SOLANA_MAINNET_USDC_MINT } from "./settlement-plan";
 import {
   assertFinalizedUsdcFundingTransfer,
   assertFinalizedUsdcTransfer,
+  assertSettlementChronology,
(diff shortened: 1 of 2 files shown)


lalalune:

PR #347, opened 2026-09-04
fix(identity): verify OAuth flow insertion

hermesagent270-commits:
## Summary

- require D1 to report both success and exactly one inserted row before exposing a new OAuth flow
- return the existing `flow_unavailable` response when D1 reports a non-throwing failure or zero-row write
- add persistence regressions for both failed postconditions

## Bug

`createFlow` returned `true` for every D1 call that did not throw. A D1 result with `success: false` or `meta.changes: 0` therefore caused the public endpoint to return an authorization URL and polling capability for a flow that was never persisted. The client could only poll a dead flow afterward.

## Verification

- `bunx vitest run workers/identity/persistence.test.ts workers/identity/identity.test.ts workers/identity/rate-limit.test.ts` (25 passed)
- `bun run typecheck`
- `bunx biome check workers/identity/persistence.ts workers/identity/persistence.test.ts`
- `git diff --check`

### DIFF
--- a/workers/identity/persistence.test.ts
+++ b/workers/identity/persistence.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it } from "vitest";
+import type { OAuthFlow } from "./contracts";
+import { type D1Database, D1IdentityPersistence } from "./persistence";
+
+const flow: OAuthFlow = {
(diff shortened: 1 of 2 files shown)


lalalune:

PR #348, opened 2026-09-04
fix(identity): verify assertion persistence

hermesagent270-commits:
## Summary

- require the exact one-time assertion row to exist in D1 before marking its OAuth flow issued
- fail with `503 assertion_unavailable` while leaving the completed flow retryable when persistence fails
- verify both the handler behavior and D1 postcondition

## Bug

The poll handler ignored the result of assertion persistence and immediately marked the OAuth flow consumed. A non-throwing D1 failure could therefore return an assertion that had no database row; the trace API would reject it, and polling again returned `410` because the flow was already marked issued.

The persistence boundary now confirms the exact stored identity, audience, timestamps, and unconsumed state before issuance proceeds. This also safely supports retry after an insert succeeded but the caller lost its response.

## Verification

- `bunx vitest run workers/identity/persistence.test.ts workers/identity/identity.test.ts workers/identity/rate-limit.test.ts` (27 passed)
- `bun run typecheck`
- focused Biome check
- `git diff --check`

### DIFF
--- a/workers/identity/contracts.ts
+++ b/workers/identity/contracts.ts
@@ -56,7 +56,9 @@ export interface IdentityPersistence {
     pollCapabilityHash: string,
     now: string,
   ): Promise<OAuthFlow | null>;
-  createAssertion(assertion: IdentityAssertion): Promise<boolean>;
+  createAssertion(
+    assertion: IdentityAssertion,
+  ): Promise<IdentityAssertion | null>;
   markAssertionIssued(flowId: string, issuedAt: string): Promise<boolean>;
   consumeAssertion(
     tokenHash: string,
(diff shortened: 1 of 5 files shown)


lalalune:
Fixed the confirmed retry regression at exact head `176a25972b0843d8a55192f49cdf84a6a2353c23`.

`createAssertion` now returns the durable assertion row after the idempotent insert. A retry with the deterministic token validates actor/login/audience/unconsumed state, retains the original creation and expiry timestamps, and returns that original expiry to the client. The handler also rejects malformed, future-created, or expired durable rows. This avoids both duplicate issuance and the prior permanent 503 after a committed insert followed by a downstream failure.

Added regressions for:
- a retry two seconds later returning the original persisted assertion rather than rejecting timestamp differences;
- a simulated post-insert flow-update failure, followed by successful recovery with the original expiry and one stored assertion.

Local evidence:
- identity handler + persistence: 15/15 pass
- TypeScript no-emit check: pass
- Biome on all five touched files: clean
- `git diff --check`: clean

Hosted exact-head checks are running; the PR remains draft pending complete validation.

lalalune:

PR #349, opened 2026-09-04
fix(wallet): reject future claim observations

hermesagent270-commits:
## Summary

- reject operator wallet-claim imports whose observation time is later than the server's creation time
- use one captured server timestamp for the claim and its audit record
- add a regression proving future provenance cannot create public claim state

## Bug

The operator recovery endpoint validated only the syntax of `observedAt`. It accepted a future timestamp, hashed it into the immutable claim, and published it as wallet provenance even though that observation had not happened yet. Future-dated records can also distort proposal cutoffs that rely on observation time.

## Verification

- `bunx vitest run functions/api/v1/trace-api.test.ts` (30 passed)
- `bun run typecheck`
- `bunx biome check backend/trace/handler.ts functions/api/v1/trace-api.test.ts`
- `git diff --check`

### DIFF
--- a/backend/trace/handler.ts
+++ b/backend/trace/handler.ts
@@ -354,6 +354,10 @@ async function createFallbackWalletClaim(
     fail(400, "invalid_request", "Invalid Solana address");
   }
   const observedAt = requiredString(body, "observedAt", validIsoTimestamp);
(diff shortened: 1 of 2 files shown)


lalalune:

PR #350, opened 2026-09-04
fix(trace): verify upload intent insertion

hermesagent270-commits:
## Summary

- require D1 to report a successful one-row insert before returning a new trace upload intent
- fail instead of handing a contributor a capability with no durable backing row
- add regressions for both explicit D1 failure and zero-row success results

## Bug

`createUploadIntent` returned `{ status: "created" }` whenever `.run()` did not throw. If D1 returned `success: false` or `meta.changes: 0`, the API still exposed an upload URL whose capability could never resolve; the next upload returned used-or-expired even though the client had just received it.

## Verification

- `bunx vitest run functions/api/v1/cloudflare-persistence.test.ts functions/api/v1/trace-api.test.ts` (37 passed)
- `bun run typecheck`
- `bunx biome check backend/trace/cloudflare-persistence.ts functions/api/v1/cloudflare-persistence.test.ts`
- `git diff --check`

### DIFF
--- a/backend/trace/cloudflare-persistence.ts
+++ b/backend/trace/cloudflare-persistence.ts
@@ -552,8 +552,9 @@ export class CloudflareTracePersistence implements TracePersistence {
         ? { status: "existing", value: mapped }
         : { status: "conflict" };
     }
+    let inserted: D1Result;
     try {
-      await this.db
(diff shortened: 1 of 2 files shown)


lalalune:

PR #351, opened 2026-09-04
fix(trace): reject future progress events

hermesagent270-commits:
## Summary

- reject contributor progress events more than five minutes ahead of server time
- capture one server timestamp for both validation and event creation
- add a regression proving a future `run_completed` event never reaches persistence

## Bug

The trace API validated only the syntax of agent-supplied `occurredAt`. A contributor could attach events arbitrarily far in the future, including `merged` or `run_completed`, creating lifecycle metadata that contradicts the run and cannot yet have occurred.

The five-minute tolerance preserves bounded client clock skew while rejecting fabricated future chronology.

## Verification

- `bunx vitest run functions/api/v1/trace-api.test.ts` (30 passed)
- `bun run typecheck`
- `bunx biome check backend/trace/handler.ts functions/api/v1/trace-api.test.ts`
- `git diff --check`

### DIFF
--- a/backend/trace/handler.ts
+++ b/backend/trace/handler.ts
@@ -46,6 +46,7 @@ export type TraceApiDependencies = {
 type ApiError = Error & { status?: number; code?: string };
 
 export const TRACE_API_CONTRACT_VERSION = "private-trace-v1-opaque-hmac-v1";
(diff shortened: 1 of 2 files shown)


lalalune:

PR #352, opened 2026-09-04
fix: reject impossible trace timestamps

hermesagent270-commits:
## Problem

The shared ISO timestamp validator relies on Date.parse after checking the string shape. JavaScript normalizes impossible dates such as February 30, so the trace API accepts and persists a timestamp that never existed. The validator also protects wallet fallback observations and intake attestations.

## Fix

Require the parsed instant to serialize back to the submitted timestamp, while preserving the supported form that omits .000 milliseconds.

## Verification

- Reproduced before the fix: impossible progress-event date returned 201
- bunx vitest run functions/api/v1/trace-api.test.ts (30 passed)
- bun run typecheck
- bunx biome check backend/trace/validation.ts functions/api/v1/trace-api.test.ts

### DIFF
--- a/backend/trace/validation.ts
+++ b/backend/trace/validation.ts
@@ -78,11 +78,20 @@ export function validEventKind(value: unknown): value is RunEventKind {
 }
 
 export function validIsoTimestamp(value: unknown): value is string {
-  return (
-    typeof value === "string" &&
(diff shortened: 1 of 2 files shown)


lalalune:

PR #356, opened 2026-09-04
fix: prevent orphan traces at upload expiry

hermesagent270-commits:
## Problem

The upload handler checked capability expiry before reading the body, but generated a new timestamp only after writing the immutable R2 object and used that later time in the atomic D1 attachment. A request authorized just before expiry could therefore:

1. pass the capability check;
2. write permanent trace bytes to R2;
3. cross the expiry boundary;
4. fail D1 attachment with 409.

That leaves an untracked immutable object even though the upload was rejected.

## Fix

Capture one authorization timestamp when the capability is resolved and use that same timestamp for atomic intent consumption. Capability validity can no longer change midway through one bounded request. Validation failures and storage failures still leave the capability retryable, and concurrent consumption remains guarded by D1.

## Reproduction

The regression sets expiry one millisecond after authorization and advances the clock while the upload is processed. Before the fix the API returned 409 after retaining bytes; it now completes the already-authorized upload and records its attachment.

## Verification

- bunx vitest run functions/api/v1/trace-api.test.ts (30 passed)
- bun run typecheck
- bunx biome check backend/trace/handler.ts functions/api/v1/trace-api.test.ts

PR #340 changes client retry behavior for transient 5xx responses. It does not address this server-side expiry race.

### DIFF
--- a/backend/trace/handler.ts
+++ b/backend/trace/handler.ts
@@ -621,9 +621,10 @@ async function uploadTraceCapability(
     fail(404, "not_found", "Upload capability not found");
   }
   const tokenHash = await sha256Hex(new TextEncoder().encode(capability));
+  const authorizedAt = deps.now().toISOString();
   const intent = await deps.persistence.getUploadIntent(
     tokenHash,
-    deps.now().toISOString(),
+    authorizedAt,
   );
   if (intent === null) {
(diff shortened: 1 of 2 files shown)


lalalune:

PR #357, opened 2026-09-04
fix: reject funding transaction reuse across ledgers

hermesagent270-commits:
## Problem

The consolidated funding index rejects one transaction attributed to multiple projects inside the direct-funding ledger and does the same independently inside the commitment ledger. It never compares the two ledgers, however, so the same network transaction can be accepted once as verified direct funding and again as a verified commitment deposit.

That gives one transfer two incompatible accounting classifications and permits duplicated recorded value across the public funding surfaces.

## Fix

Reject any network/transaction identity already present in direct funding when processing commitment records. This extends the index's existing one-transaction/one-attribution invariant across its ledger boundary.

## Reproduction

The regression constructs one finalized Solana USDC transfer to a declared Squads vault and records it as both direct project funding and a commitment deposit. The index previously accepted both records; it now fails closed.

## Verification

- bunx vitest run src/lib/funding.test.ts src/lib/funding-commitment.test.ts src/lib/project-schema.test.ts (34 passed)
- bun run funding:check
- bun run typecheck
- bunx biome check src/lib/funding.ts src/lib/funding.test.ts

### DIFF
--- a/src/lib/funding.test.ts
+++ b/src/lib/funding.test.ts
@@ -3,6 +3,7 @@
 import { describe, expect, it } from "vitest";
 import {
   assertProjectFundingAddresses,
+  assertProjectFundingIndex,
   assertProjectFundingLedger,
   assertProjectFundingRecord,
   currentProjectFundingRecords,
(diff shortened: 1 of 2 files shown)


lalalune:

PR #358, opened 2026-09-04
fix: reject noncanonical Solana transaction signatures

hermesagent270-commits:
## Problem

Settlement evidence, reward-settlement manifests, the Solana RPC client, and the core USDC reconciler accepted any 64-128 character base58-looking string as a transaction signature. Solana signatures are exactly 64 decoded bytes; for example, 64 repeated base58 characters decode to fewer bytes but passed every loose check.

With an RPC-shaped response containing the same malformed label, the core verifier accepted it as finalized transaction evidence. This also allowed impossible identifiers into public settlement records.

## Fix

Use the existing decoded-length-aware isSolanaTransactionId validator consistently at all four boundaries:

- settlement evidence ingestion
- public settlement manifest validation
- RPC request validation
- core funding and payout reconciliation

## Reproduction

Regression fixtures use a 64-character base58 string that does not decode to 64 bytes. Before the fix the RPC was called and both core verification paths accepted it; now it is rejected before network access and by the reconciler itself.

## Verification

- bunx vitest run src/lib/solana-settlement.test.ts scripts/solana-rpc.test.ts src/lib/rewards.test.ts scripts/verify-funding-solana.test.ts scripts/money-state-commands.test.ts (30 passed)
- bun run typecheck
- Biome checks on all seven changed files

PR #355 fixes the distinct requirement that a valid signature must be first in the transaction signature array. This PR validates its decoded byte length and may need a small rebase in the shared reconciler.

### DIFF
--- a/scripts/solana-rpc.test.ts
+++ b/scripts/solana-rpc.test.ts
@@ -54,6 +54,19 @@ describe("Solana RPC boundary", () => {
     ).rejects.toThrow("exceeded its size limit");
   });
 
+  it("rejects a base58 value that is not a 64-byte signature", async () => {
+    const fetcher = vi.fn(async () => rpcResponse({ slot: 42 }));
+
+    await expect(
+      fetchFinalizedSolanaTransaction(
+        "https://api.mainnet-beta.solana.com",
+        "2".repeat(64),
+        { fetcher },
+      ),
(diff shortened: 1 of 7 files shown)


lalalune:

PR #360, opened 2026-09-04
fix: bind Sablier commitments to the project recipient

hermesagent270-commits:
## Problem

A reviewed Sablier commitment instrument contains the Lockup contract and stream ID but no recipient. The live verifier checks a recipient supplied on the command line, while the durable commitment record matches only contract and stream ID. Nothing therefore proves that a recorded stream pays the project whose committed pool it backs.

A project could reference a real USDC stream payable to another address and count its locked balance toward committedMinor.

## Fix

- require a canonical recipient in every Sablier instrument
- include that recipient in durable commitment-record identity and correction identity
- require records to match the exact manifest-bound recipient
- continue treating contract plus stream ID as the unique on-chain instrument, so declaring the same stream with two recipients is rejected rather than creating two identities

No current project manifest or funding record declares a Sablier commitment, so this tightens the pre-activation schema without migrating published evidence.

## Verification

- regression accepts the exact manifest recipient and rejects another valid EVM address
- duplicate stream declarations with different recipients fail closed
- bunx vitest run src/lib/funding-commitment.test.ts src/lib/project-schema.test.ts (26 passed)
- bun run projects:check
- bun run funding:check
- bun run typecheck
- biome checks pass

PR #354 fixes a distinct RPC quorum-state issue and does not add this durable manifest binding.

### DIFF
--- a/src/lib/funding-commitment.test.ts
+++ b/src/lib/funding-commitment.test.ts
@@ -19,6 +19,7 @@ const FUNDER_MEMBER = "Stake11111111111111111111111111111111111111";
 const STEWARD_MEMBER = "SysvarRent111111111111111111111111111111111";
 const DEPOSIT_SIGNATURE = "3".repeat(88);
 const RELEASE_SIGNATURE = "4".repeat(88);
+const SABLIER_RECIPIENT = `0x${"a".repeat(40)}`;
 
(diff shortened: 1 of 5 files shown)


lalalune:

PR #361, opened 2026-09-04
fix(trace): reject events racing finalization

hermesagent270-commits:
## Problem

The event handler checked that a run was not finalized, then inserted the event in a separate persistence call. Finalization could commit between those operations, allowing a contributor progress event to be appended after the run became finalized despite the API contract rejecting events on finalized runs.

## Fix

Make event insertion conditional on an actor-owned run still being non-finalized in the same D1 statement. Treat a zero-row insert as a conflict, closing the time-of-check/time-of-use window and also failing closed for missing or mismatched runs.

## Reproduction

The persistence regression models a run that finalized after the handler read it. Before the fix, a non-throwing zero-row database result was reported as a created event. The regression now requires a conflict and verifies the atomic run-state predicate.

## Verification

- focused persistence and trace API tests: 36 passed
- full Vitest pass: 649 passed; 19 sandbox child-git failures plus two environment-only suite load failures (missing PyYAML and node:sqlite bundling)
- bun run typecheck
- Biome check on both changed files
- git diff --check

No open PR addresses this finalized-run event insertion race.

### DIFF
--- a/backend/trace/cloudflare-persistence.ts
+++ b/backend/trace/cloudflare-persistence.ts
@@ -499,12 +499,14 @@ export class CloudflareTracePersistence implements TracePersistence {
         : { status: "conflict" };
     }
     try {
-      await this.db
+      const inserted = await this.db
         .prepare(
           `INSERT INTO run_progress_events (
             id, run_id, github_user_id, kind, occurred_at, source,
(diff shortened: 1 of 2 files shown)


lalalune:

PR #362, opened 2026-09-04
fix(wallet): refresh claims after GitHub login rename

hermesagent270-commits:
## Problem

Wallet claims are correctly owned by the stable numeric GitHub actor ID, but an unchanged-address registration returned the current claim solely by comparing wallet addresses. After a GitHub username rename, that returned the immutable claim containing the old login.

Reward preparation independently compares the registry login with GitHub's current login and rejects the stale claim. The contributor therefore could not refresh valid payout provenance while keeping the same wallet address.

## Fix

Treat an unchanged wallet registration as idempotent only when both the address and login (case-insensitively) still match. When the stable actor ID is unchanged but the login changed, append a same-address successor containing the current authenticated login.

Case-only username spelling remains idempotent, and the existing predecessor/concurrency checks continue to protect the single claim lineage.

## Verification

- regression proves the same numeric actor can refresh the same address after a login rename
- bunx vitest run functions/api/v1/trace-api.test.ts scripts/github-wallets.test.ts (37 passed)
- bun run typecheck
- bunx biome check backend/trace/handler.ts functions/api/v1/trace-api.test.ts
- git diff --check

No open wallet PR addresses login-renamed claim refresh; #343 covers claim/audit atomicity, #349 future observations, and #335 proposal cutoffs.

### DIFF
--- a/backend/trace/handler.ts
+++ b/backend/trace/handler.ts
@@ -265,7 +265,10 @@ async function createContributorWalletClaim(
       "Wallet claim changed; reload the current claim before submitting",
     );
   }
-  if (current?.walletAddress === walletAddress) {
+  if (
+    current?.walletAddress === walletAddress &&
+    current.githubLogin.toLowerCase() === actor.githubLogin.toLowerCase()
+  ) {
     return json(200, publicWalletClaim(current));
   }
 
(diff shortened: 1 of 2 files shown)


lalalune:

PR #363, opened 2026-09-04
fix(cycles): reject files ahead of lifecycle state

hermesagent270-commits:
## Problem

The browser-facing cycle-index validator required later states to have their prerequisite files, but did not enforce the reverse direction. A review entry could reference an allocation or execution plan, and a payment-ready entry could already reference an execution plan, while still passing validation.

Those combinations cannot be emitted by the canonical cycle builder and expose contradictory public money state: the lifecycle label says no approval or settlement plan exists while its immutable-file links say otherwise.

## Fix

Make the file/state mapping bidirectional:

- review permits no allocation, execution plan, or settlement
- payment-ready requires its allocation but permits no execution plan or settlement
- existing settlement-planned behavior remains unchanged so failed or partial settlement records remain representable

## Verification

- regressions cover allocation during review, plan during review, and plan during payment-ready
- bunx vitest run src/lib/cycle-index.test.ts (11 passed)
- bun run typecheck
- bun run cycles:check (validated the published reward cycle)
- bunx biome check src/lib/cycle-index.ts src/lib/cycle-index.test.ts
- git diff --check

No open PR overlaps this cycle-index file/state validation gap.

### DIFF
--- a/src/lib/cycle-index.test.ts
+++ b/src/lib/cycle-index.test.ts
@@ -170,6 +170,57 @@ describe("public cycle index", () => {
     );
   });
 
+  it("rejects lifecycle files that are ahead of the declared state", () => {
+    const allocationDuringReview = entry();
+    allocationDuringReview.files.allocation = {
+      sha256: DIGEST,
+      url: "/data/cycles/eliza/2026-07/allocation.json",
+    };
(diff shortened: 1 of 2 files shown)


lalalune:

PR #364, opened 2026-09-04
fix(policy): bind terms to their activation date

hermesagent270-commits:
## Problem

Active project terms declare that they are non-retroactive and maintain an ordered immutable receipt-policy binding history. The schema binds the latest entry to the current revision and source hashes, but not to terms.effectiveAt.

A well-shaped manifest could therefore retain the exact same current binding while moving the declared effective date into the past or future. That makes the public terms chronology disagree with the activation record that is supposed to anchor it.

## Fix

Require active terms.effectiveAt to equal the latest receipt-policy binding's activatedAt. Pending policies remain unchanged because they do not yet have an activation binding.

All current project manifests already satisfy this invariant.

## Verification

- regression rejects a retroactively shifted effective date with otherwise unchanged binding evidence
- synthetic activation fixture now declares the same effective and activation instant
- bunx vitest run src/lib/project-schema.test.ts scripts/check-project-transitions.test.ts (24 passed)
- bun run projects:check
- bun run typecheck
- bunx biome check src/lib/project-schema.mjs src/lib/project-schema.test.ts
- git diff --check

No open PR overlaps project terms effective-date binding.

### DIFF
--- a/src/lib/project-policy.test.ts
+++ b/src/lib/project-policy.test.ts
@@ -50,6 +50,7 @@ describe("project policy transitions", () => {
     authority: unknown;
     terms: {
       revision: string;
+      effectiveAt: string;
       repositoryLicense: { fileSha256: string | null };
       inbound: { fileSha256: string | null };
       receiptPolicy: {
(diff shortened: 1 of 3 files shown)


lalalune:
Repaired the failing policy fixtures at exact head `0bf0c9a5275968a395f06f93e4577cbb47648c58`.

The active fixture now binds `terms.effectiveAt` to its latest receipt-policy activation, and the multi-append negative case advances `effectiveAt` to the final proposed binding. This preserves the intended transition assertions instead of weakening error expectations.

Local evidence:
- project-policy + project-schema: 26/26 pass
- Biome on touched test: clean
- `git diff --check`: clean

Hosted exact-head checks are running; the PR remains draft pending complete validation.

lalalune:

PR #365, opened 2026-09-04
fix(funding): bind index timestamp to its ledger

hermesagent270-commits:
## Problem

The funding-index builder deterministically sets generatedAt to the latest observedAt across direct-funding and commitment records. The public validator only checked that generatedAt was a timestamp, however, and accepted a stale value or null alongside non-empty records.

That allows the published as-of time to contradict the ledger it describes and weakens freshness checks by separating aggregate metadata from its source records.

## Fix

Recompute the maximum observation timestamp from both validated ledgers and require generatedAt to equal it exactly. An empty index still requires null, matching the builder's existing behavior.

## Verification

- regression accepts the exact latest observation and rejects both a stale timestamp and null
- bunx vitest run src/lib/funding.test.ts src/lib/funding-commitment.test.ts (16 passed)
- bun run funding:check
- bun run typecheck
- bunx biome check src/lib/funding.ts src/lib/funding.test.ts
- git diff --check

PR #357 touches the same two files for the distinct cross-ledger transaction-reuse invariant. This change has no semantic overlap and may need a small rebase if that PR merges first.

### DIFF
--- a/src/lib/funding.test.ts
+++ b/src/lib/funding.test.ts
@@ -3,6 +3,7 @@
 import { describe, expect, it } from "vitest";
 import {
   assertProjectFundingAddresses,
+  assertProjectFundingIndex,
   assertProjectFundingLedger,
   assertProjectFundingRecord,
   currentProjectFundingRecords,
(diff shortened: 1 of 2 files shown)


lalalune:

PR #359, opened 2026-09-04
fix(skills): serialize concurrent live-report discovery

rndrntwrk:
## Summary

- serialize complete live-report discovery under one cross-process lock keyed by the authenticated `github.com` identity, shared by the Eliza, ASI, Delta Star, and Heir reporters even across repositories and `TMPDIR` values
- make contention an explicit exit status `2` before discovery, then re-probe the direct GraphQL budget only after lock acquisition
- prevent split-lock, PID-reuse, and interrupted-owner takeovers with atomic directory ownership plus process-birth and in-flight child leases; reject malformed, symlinked, wrong-owner, or over-permissive lock evidence fail-closed
- preserve the existing exact-head and publication authorization gates

Closes #329

## Validation

- `bun run test` — passed on `6e3c9bbd7e550cb290dce0f5b209451dbad5486a`; 67 Vitest files / 680 tests and 112 Bun skill tests passed
- `bun run typecheck`, `bun run format:check`, and `bun run lint:check` — passed on the exact head; lint retained 50 pre-existing `src/styles.css` specificity warnings
- `bun run test:e2e` — passed on the exact head after preparing the repository's documented pull-request ledger fixture from `https://slop.cash/data/leaderboard.json`; 36 desktop/mobile browser tests and 1 Pages artifact-contract test passed, including the production build
- focused live-report suite — 56 passed under Node `v24.15.0`, including 9 serialization/security/recovery cases
- Windows filtered-job reproduction — the exact receipt test passed with `SLOP_TEST_NODE` deliberately set to a nonexistent executable, proving filtered receipt coverage does not initialize the live-report fixture
- exact-head GitHub CI — all required checks passed on `6e3c9bbd7e550cb290dce0f5b209451dbad5486a`, including the dependency audit, Windows lifecycle, project/skill validation, unit tests, build, and browser matrix. The first CI run's external audit request timed out; the replacement exact-head run completed it successfully.
- independent exact-head review — OpenAI GPT-5.6 Sol at xhigh reasoning reported no findings; base `a2fe88318864e11c72cfa96cbfe4215a166e608b` and head `6e3c9bbd7e550cb290dce0f5b209451dbad5486a` were unchanged before and after review

## Evidence

- Before screenshots: N/A - CLI-only concurrency defect; no UI changed.
- After screenshots: N/A - CLI-only concurrency defect; no UI changed.
- Walkthrough video: N/A - behavior is exercised by deterministic concurrent-process tests.
[...]

PR #337, opened 2026-09-03
fix(site): never print a pledged pool as a dollar pool

drew-dot-com:
Part (a) of #331. Parts (b) (allocator follows `committedMinor`) and (c) (trial month, then `$0`) stay open in the issue; this PR does not close it.

## What

Every monthly-pool surface printed the cap as money while all four projects sit at `fundingState: "pledged"` with `committedMinor: "0"`: `$10,000 / month` on project cards, `$10,000 monthly pool` on the leaderboard tabs and selected-cycle label, `$10,000` in the home proof object, and `$10,000` on the project reward card. A pool with nothing behind it now reads **Unfunded**, with the cap shown as a target. The dollar figure returns only once `fundingState` is `committed` and `committedMinor` is positive.

| Surface | Before | After |
| --- | --- | --- |
| Project card | `$10,000` / month · "Projected cap · funding uncommitted" | `Unfunded` · target $10,000 / month · "Target cap · no funding committed" |
| Leaderboard tab and label | `$10,000 monthly pool` | `unfunded, target $10,000` |
| Home proof object | `$10,000` · Type / Funding / Payment hardcoded | `Unfunded` · Target `$10,000 monthly cap` · Funding and Payment read from the manifest |
| Project reward card | `$10,000` · "Up to this amount is allocated each month." | `Unfunded` · "Target $10,000 per month. No funding is committed, so no payment is scheduled until the project commits funds." |

## How

- `src/App.tsx`: two exported helpers, `monthlyPoolUnfunded(reward)` (`fundingState !== "committed" || committedMinor === 0`) and `monthlyPoolLabel(reward)`, used at the four sites above. Committed pools keep every existing string. No CSS changes; the existing card and reward-card layouts carry the word without overflow at 1440 and 390.
- Copy only. The leaderboard "simulated share" column, the cycle `Suggested` figure, and the contributor-page projection still derive from the full cap through `project-view.ts`; that is (b) in #331.

## Tests

- `tests/app.test.tsx`: unit test for `monthlyPoolLabel` (pledged, committed-with-zero, committed-with-balance); home card, leaderboard tab, and project reward card assertions updated to the new strings and to assert the old `$10,000 monthly pool` string is gone.
- `tests/e2e/site.spec.ts`: card and tab assertions updated the same way.
- `bun x vitest run`: 681 pass across 67 files. `bun run typecheck`, `bun run lint:check`, `bun run format:check`: pass.
[...]

PR #372, opened 2026-09-05
Resolve reviewed receipt, funding, wallet, and scan issues

### LINKED ISSUE #328: Clarify whether additive review-budget allocation includes finalized-trace weight
## Context

PR #322 implemented additive review-budget accounting while preserving each review event in the existing shared-pool calculation. The shared-pool projection weights an event with its `evidenceBonusBasisPoints`, consistent with Score v2 granting additional weight for eligible finalized trace evidence.

The dedicated review line appears to aggregate raw review `scoreThirds` instead. As a result, two otherwise equal review events can receive different shared-pool weights based on finalized trace evidence while remaining equal within the additional review-budget allocation.

This boundary was raised in https://github.com/SlopDotCash/slopdotcash/pull/322#issuecomment-5512952906 after merge, but it is not currently tracked by an open issue.

## Why this needs an explicit rule

Either policy can be coherent:

- Evidence-adjusted allocation would apply the existing Score v2 weight consistently to both funding lines.
- Tier-only allocation would make the dedicated line reward substantive review tier while intentionally ignoring trace evidence.

The current implementation and protocol do not make that distinction sufficiently explicit. This is a policy ambiguity before activation, not a claim that a funded payout has already been miscalculated. No project currently has an active funded review line.

## Suggested resolution

Please choose and document one invariant before a review budget is funded:

1. Use the same evidence-adjusted review weight as the shared pool, with a focused allocation test showing that an eligible 15% finalized-trace bonus changes the dedicated-line allocation.
2. Keep the dedicated line tier-only, explicitly state that `evidenceBonusBasisPoints` applies only to the shared-pool side, and add a focused test showing that trace evidence does not change dedicated-line allocation.

A two-review fixture with equal `scoreThirds` and only one eligible finalized-trace bonus should make the selected behavior unambiguous.

### LINKED ISSUE #329: Serialize concurrent live-report scans that share one GraphQL budget
## Summary

PR #317 made each canonical `live-report.mjs` probe the direct GraphQL `rateLimit` before choosing its activity-discovery path. That fixes the stale REST preflight from #314, but the affordability decision remains process-local.

Multiple agents and all project skills can use the same authenticated GitHub account and therefore the same GraphQL bucket. There is no cross-process lock or budget reservation around complete activity discovery. Two reports can both observe an affordable budget, both enter the expensive scan, and jointly exhaust the bucket. A prompt-level convention that names one coordinator does not enforce mutual exclusion across independent agent processes.

## Observed overlap

On 2026-09-03, one host had these reports overlapping under the same authenticated account:

- ASI report started at 09:07:37 MSK
- Eliza report started at 09:08:09 MSK
- another Eliza report started at 09:08:44 MSK before the prior Eliza report completed

During the overlap, direct GraphQL samples moved from 3,601 remaining to 3,487 and then 2,974. These samples prove concurrent consumption but do not attribute the entire delta to a particular process.

The activity scan is materially expensive because it paginates the open queue and requests pull comments, reviews, review threads, and nested review-thread comments. The canonical threshold is currently 1,000 points, so independent reports can each pass their preflight while their combined work exceeds the shared remainder.

An older installed skill revision can amplify this by still trusting the stale REST GraphQL value, but that is not required to reproduce the concurrency race. The current canonical script at develop `a2fe88318864e11c72cfa96cbfe4215a166e608b` contains the direct probe and still has no shared scan lock.

## Expected behavior

Complete activity discovery should be globally serialized for reports using the same local GitHub identity:

1. Acquire a cross-process lock shared by the ASI, Delta Star, Eliza, and Heir report implementations before the expensive discovery phase.
2. Re-probe the direct GraphQL budget after acquiring the lock, rather than relying on the value observed while waiting.
3. If another report holds the lock, wait for a bounded interval or return an explicit non-success status without starting discovery.
4. Release the lock on success, handled failure, interruption, and process termination without an unlink-and-recreate split-lock race.
5. Keep exact-head and publication gates fail-closed. Serialization must not turn an incomplete report into authorization to write.

## Suggested acceptance coverage

- Two concurrent report processes sharing one injected identity never execute complete discovery simultaneously.
- The second process rechecks the budget after the first releases the lock and chooses GraphQL, REST, or failure from that fresh value.
- Reports for different project repositories still contend on the same shared identity lock.
- A failed or interrupted first process does not permanently block later reports.
- The emitted result distinguishes waiting or lock contention from GitHub rate-limit exhaustion.

This issue is about local scan serialization, not increasing GitHub limits or weakening the completeness requirement.

### LINKED ISSUE #332: Wallet cut-off: a wallet observed after proposal generation applies to the next cycle

### LINKED ISSUE #353: Receipt CLI cannot recover a run after a transient private-trace upload failure
A contributor running the review lane today hit `trace upload request returned HTTP 500` from the private-trace API. They correctly did not post the review. Intake was live at the time (attestation `verifiedAt` 2026-09-04T05:04:15Z on `a2fe883`) and every write route on `api.slop.cash` answers with the expected 4xx now, so this reads as a one-off internal error. That is not the bug. The bug is that one transient 500 strands the run permanently, because the receipt CLI cannot recover from it.

## What happens on retry

`run-receipt.mjs trace` on the same run id replays deterministic idempotency keys (`skills/contribute-to-eliza/scripts/run-receipt.mjs:1527`, `:1554`, `:1626`; the four project copies are byte-identical).

1. `POST /api/v1/runs` replays fine and returns the existing run.
2. `POST /api/v1/runs/{id}/trace-intents` uses the key `intent:<serverRunId>:<sha256>`. `createUploadIntent` returns the original intent, including its original `expiresAt`, and the HMAC-derived `uploadUrl` is the same capability.
3. The intent lived five minutes. Any retry after that hits `getUploadIntent`'s `expires_at > ?` guard and the `PUT` returns `410 upload_capability_invalid`.

There is no client path that mints a fresh intent for an existing run. `backend/trace/README.md` says "A failed, expired, or already-used capability cannot be retried; the authenticated client creates a new intent." The client cannot.

A second dead end: if the bytes attached and the 500 came from finalization, the run is now `trace_uploaded`, and the client refuses at step 1 because it requires `created.state === "awaiting_trace"` (`run-receipt.mjs:1536`). No resume path exists.

The only recovery today is to start a new run and redo the work, which also discards the receipt already emitted for that run.

## Proposed fix

Client side only, no API change:

- Salt the intent idempotency key with an attempt counter or the client timestamp, so a retry after expiry creates a new intent. The server already tolerates this: the R2 object is content-addressed and `attachTrace` accepts a matching digest on a run that has none yet.
- Accept `trace_uploaded` from run creation and skip straight to finalize when the reported `traceSha256` matches the local digest.
- Persist `serverRunId` and the last completed step in the active run state so a retry knows where it stopped.

## Related

`backend/trace/handler.ts:1050` logs only `private trace API failure` for every 5xx and drops the thrown message. The thrown messages are static strings, so logging `error.message` there leaks nothing and would have said which of R2, D1, or the identity binding failed. Without it this class of failure is undiagnosable even from the Cloudflare side.

Issue only. Happy to open the client PR if the salting approach is acceptable.

lalalune:
## Summary

Integrates reviewed contributor PRs #327, #330, #335, #336, #337, #339, #342 and #359, preserving their commit ancestry, with integration repairs:

- Retry changing GitHub inventories with a fresh budget under one identity-scoped scan lock.
- Retain owner-specific stale-lock tombstones so delayed reapers cannot displace a successor's live lock.
- Recover uploaded/finalized receipt runs; a new authenticated retry requests a fresh upload capability instead of renewing an expired capability.
- Enforce the proposal-generation wallet cutoff.
- Reject cancelable Sablier streams as committed funding.
- Show actual committed amounts separately from funding targets, and explicitly explain absent receiving routes.
- Document and regression-test the existing tier-only additive review-budget allocation.
- Repair the September UI fixture whose work-item timestamps advanced beyond its fixed snapshot time, breaking scheduled CI.

Closes #328
Closes #329
Closes #332
Closes #353

Related: #302, #331, #333, #338. This does not implement allocator/trial-month policy, independent-vault stewardship, or new automated funding approval authority.

## Verification

In progress on this integration head. Focused UI/reward/Sablier tests: 75 passed. Skill/CLI suite: 58 passed, including deterministic delayed-reaper regression. Typecheck passed.

Full verification, four-size browser matrix, and hosted exact-head evidence will be added before merge. No production deployment or payout availability is claimed.

## Boundaries

- No shared-checkout changes, secrets, private traces, addresses, or payment credentials added.
- No signing, broadcasting, approval bypass, environment, or deployment configuration changes.
- Existing immutable cycle records are unchanged.
- Stale scan-lock tombstones contain local process metadata only and are intentionally retained for safe concurrent recovery.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -173,9 +173,11 @@ shared pool unchanged; a committed review line may pay them only on top of that
 treatment, with separate public arithmetic and evidence.
 
 Proposal review lasts 14 days. Project owners may adjust awards within the cap
-with a public reason. Wallet changes append a successor and reset review;
-history is never edited. Missing wallets remain unclaimed. Related-party money
-requires separate approval.
+with a public reason; only such amount changes reset review. Wallet changes
+append a successor; a wallet observed after proposal generation applies to the
+next cycle and never touches the current proposal. History is never edited.
+Missing wallets remain unclaimed. Related-party money requires separate
+approval.
 
 Settlement tools create unsigned Solana mainnet USDC plans only. `paid`
 requires finalized evidence whose exact source and destination deltas reconcile
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -173,9 +173,11 @@ shared pool unchanged; a committed review line may pay them only on top of that
 treatment, with separate public arithmetic and evidence.
 
 Proposal review lasts 14 days. Project owners may adjust awards within the cap
-with a public reason. Wallet changes append a successor and reset review;
-history is never edited. Missing wallets remain unclaimed. Related-party money
-requires separate approval.
+with a public reason; only such amount changes reset review. Wallet changes
+append a successor; a wallet observed after proposal generation applies to the
+next cycle and never touches the current proposal. History is never edited.
+Missing wallets remain unclaimed. Related-party money requires separate
+approval.
 
 Settlement tools create unsigned Solana mainnet USDC plans only. `paid`
 requires finalized evidence whose exact source and destination deltas reconcile
--- a/cycles/README.md
+++ b/cycles/README.md
@@ -16,9 +16,13 @@ untouched; a directory containing only one required file is refused as partial.
 For a platform-funded monthly pool, the creator edits `proposal.json` during
 the 14-day review. The creator may set any contributor amount, including zero,
 or raise it above the deterministic suggestion while the cycle total remains
-within the published cap. Every changed amount records a public reason. Wallet
-changes update `review.lastMaterialChangeAt` and reset `review.endsAt`. The
-normal progression then adds, without replacing earlier files:
+within the published cap. Every changed amount records a public reason and
+updates `review.lastMaterialChangeAt`, which resets `review.endsAt`. Wallets
+are cut off at `generatedAt`: a wallet observed after the proposal was
+generated applies to the next cycle and never modifies the current proposal
+or its review clock. The row stays `unclaimed` and carries forward. Only a
+creator amount change moves `review.lastMaterialChangeAt`. The normal
+progression then adds, without replacing earlier files:
 
 Transfers have a 2 USDC minimum. Smaller awards remain
 `held-below-minimum`, retain their exact integer micro-USDC amount, and accrue
--- a/funding/README.md
+++ b/funding/README.md
@@ -100,7 +100,7 @@ mechanism that Slop does not control: an autonomous Squads v4 multisig vault
 holding USDC on Solana or a Sablier Lockup v4 USDC stream on Base or Ethereum.
 A Squads commitment requires an exact 2-of-2 funder and project-steward
 multisig with no configuration authority; a Sablier commitment uses a
-non-upgradeable stream. Slop holds no key, admin, or fee position in any
+non-upgradeable, non-cancelable stream. Slop holds no key, admin, or fee position in any
 instrument; it publishes the reviewed reference and read-only evidence only.
 Committed funds are constrained by that reviewed third-party instrument, not
 held by Slop.
@@ -156,7 +156,7 @@ funder's claimed wallet—and are never inferred. The verifier never signs,
 broadcasts, handles a key, or writes a record.
 
 For a Sablier Lockup v4 USDC stream on Base or Ethereum, the read-only
-verifier (`commitment-sablier-v1`) queries three fixed public RPC authorities,
+verifier (`commitment-sablier-v2`) queries three fixed public RPC authorities,
 checks each authority's chain ID, pins every stream view call to that
 authority's own finalized block, and requires two to agree exactly on the
 stream state:
@@ -174,8 +174,11 @@ on-chain end time, and the canceled and depleted flags truthfully, with the
 canonical evidence URL `https://basescan.org/address/<contract>` or
 `https://etherscan.io/address/<contract>` and each agreeing authority's
 finalized block identity. Wrong chains, wrong tokens, wrong recipients, and
-malformed return data fail closed. The verifier never signs, broadcasts,
-handles a key, or writes a record.
+malformed return data fail closed. A cancelable stream also fails closed: the
+funder could reclaim the undistributed balance at any time, so it cannot back
+a positive `committedMinor`, and no evidence is emitted for it. Only a stream
+whose `isCancelable` flag is already false can be recorded. The verifier never
+signs, broadcasts, handles a key, or writes a record.
 
 A manifest may set `fundingState: "committed"` only while an active instrument
 is declared and the verified commitment ledger (deposits minus releases and
--- a/protocol/review-budget-v1.md
+++ b/protocol/review-budget-v1.md
@@ -43,6 +43,14 @@ Only accepted `substantive-review` score events participate in the additive
 line. Self-review, bot review, post-merge review, duplicates, and excluded
 evidence remain ineligible under the scoring contract.
 
+The dedicated line is tier-only: aggregate each actor's accepted review
+`scoreThirds` and allocate by deterministic largest remainder. The finalized
+trace bonus (`evidenceBonusBasisPoints`) applies to that event's shared-pool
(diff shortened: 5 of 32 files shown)


lalalune:
Local exact-head verification completed on a8ebaac588f600859e28591b1ed8659c4a5dd773: bun run verify passed, including 118 Bun skill tests; real Chromium E2E passed 36/36 across 1440, 1024, 768, and 320 pixel viewports plus the Pages artifact-contract test (1/1). Typecheck, validators, dependency audit, formatting, lint, and production build passed. Browser coverage includes axe accessibility, keyboard navigation, console/network checks, and raw Markdown/archive consistency. Public PR fixture was prepared from the deployed ledger using the checked-in script. Hosted jobs remain queued; no merge or deployment is claimed. PR #371 independently implements the same tier-only review invariant covered here; do not double-merge divergent helper implementations.

lalalune:
Concurrent-queue coordination: this integration already includes current source heads #327 135fc75, #330 79b65fa, #335 a7e5adc, #336 acfc2aa, #337 a9be8b1, #339 2225bd0, plus #342 and the original #359 with an independently tested owner-specific tombstone repair. #327 has now landed separately. Please preserve this branch while the remaining individual PRs are being merged; rebase and review its residual diff before any merge. Latest head 16a019f8e993b56b08e1414d935fbc91245a18ac passed full local verify with 123 skill tests; updated browser evidence is running. #373 contains the separately scoped upstream receiving-address gate.

lalalune:
Superseded by the individually reviewed and merged fixes (#327, #330, #335, #336, #337, #339, #342) plus the repaired scan-lock integration in #381. The old integration head also reproduced a helper/child lock-release race fixed in #381, so merging this stale conflicting aggregate would regress current develop. The tier-only review-budget work continues separately in #371.

Issue #366, opened 2026-09-04
Compromised-wallet hold: withhold an unsafe payment without forfeiting the amount or restarting review


drew-dot-com:
Follow-up to #332, prompted by @mashingaan's third point there: a wallet cut-off must not stop a creator from withholding a payment to a destination a contributor has reported as compromised. Today it does not stop the withholding, but the withholding forfeits the amount. This issue proposes the missing carry path. Issue first; PR on request.

## Today

- A creator can already set a row to `held` during review. `src/lib/rewards.ts` (around line 634) requires an `adjustmentReason` of 12 to 1000 characters whenever `approvedMinor` differs from `suggestedMinor` on a row that is not `proposed`, `unclaimed`, or `held-below-minimum`. So a hold is public, reasoned, and single-row. It does not touch `review.lastMaterialChangeAt`, so other recipients' review clock does not move.
- `held` rows do not carry. `scripts/prior-cycle-accrual.ts` (around line 153) carries only `unclaimed` and `held-below-minimum` into the next proposal. A `held` row is final for that cycle.
- The wallet on a row cannot be replaced during review. The validator (around line 878) rejects any `wallet.observedAt` newer than `lastMaterialChangeAt` today, and newer than `generatedAt` under #335. Replacing the destination means regenerating the proposal, which restarts the 14 days for everyone.

Put together: on a credible compromise report, the creator's choices are pay the reported address, hold and forfeit, or regenerate and restart every other recipient's review. None of those is right.

## Proposal

1. **A hold kind that carries.** Keep the `held` state. Add an optional `hold` object on the row:

   ```json
   "hold": {
     "kind": "unsafe-destination",
     "reportedAt": "2026-09-10T14:02:11.000Z",
     "reportedBy": { "id": 12345, "login": "contributor" }
   }
   ```

   `prior-cycle-accrual` carries a `held` row when `hold.kind === "unsafe-destination"`, using the same `accruedMinor ?? suggestedMinor` rule as `unclaimed`. All other `held` rows stay final, as today.

2. **The record keeps the original destination.** `wallet` on the held row is left exactly as generated. The audit trail shows the address that was reported, the amount, the reason, and who reported it. Nothing is rewritten.

[...]

PR #382, opened 2026-09-05
Preserve earned rewards while holding compromised wallets

lalalune:
Compromised-wallet reports can hold an unsafe destination without deleting the contributor's earned amount or restarting the cycle's review. The hold follows the wallet through later cycles, keeps shared-pool carry separate from reviewer funds, and requires the signed report to match the original wallet claim and contributor.

Signature authorization runs in the immutable trusted-base checker. Ordinary cycle validation and site packaging remain credential-free, including cycles containing held contributors during review.

Addresses #366. Final combined funding integration and browser validation remain pending.

Validation at `a4c5afa801ba16c93b1a0bdf8f633769ff6be801`: full `bun run verify` passed with 797 unit/integration tests, 124 skill tests, and production build. A clean-environment regression runs held-cycle validation, reconstruction, and build with an unavailable GitHub client. Independent review confirmed the credential boundary and hold invariants.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -143,6 +143,8 @@ jobs:
           bun scripts/prepare-private-intake-attestation.ts --ensure-disabled
 
       - name: Validate reward lifecycle
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: bun run cycles:check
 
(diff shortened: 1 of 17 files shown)


lalalune:

PR #384, opened 2026-09-05
fix: integrate funding and settlement validation safeguards

lalalune:
## Integrated funding and settlement hardening

Integrates #354, #355, #357, #358, #360, #363, #364, and #365 with their regression coverage: complete Sablier quorum state, exact recipient binding, canonical Solana transaction signatures, primary-signature transaction identity, cross-ledger replay rejection, lifecycle-file consistency, terms activation-date binding, and deterministic funding-index observation timestamps.

Updated two synthetic accessibility fixtures to declare the newly required Sablier recipient and corrected a stale verifier-version fixture. No production funding or signing authority is introduced.

Exact head 085b24ed7948dee3e056ea67382722fd5c23ca4e passed all 764 Vitest tests, schema/type/format/lint checks. Six skill checks detected the wrong local Node version; the skill suite and build are being rerun under pinned Node 24.15.0 without source changes. Actions are waived by the maintainer; neither hosted CI nor deployment is claimed. No UI changes; browser visual review is N/A for these validation-only changes.

PR #370, opened 2026-09-05
fix(ui): repair clipboard failures and project allocation drafts

lalalune:
Repair public drafting failures and misleading UI states:

- Normalize absent or denied clipboard APIs into visible copy failure feedback.
- Reset project drafts when history navigation switches projects.
- Expose allocation search whenever the ten-row display truncates the list.
- Reject draft totals above the applicable shared and funded review limits using integer micro-USDC.
- Switch navigation and footer layouts before tablet widths crowd their contents.
- Remove the fake sign-in state from the public proposal form.
- Give the CLI integration suite an explicit 60-second budget; Bun's default five-second budget killed its subprocesses.

Current head: a3a7d22bc96681f6d138e6d34f4060eb91f9d8cb. The focused component suite passes 57 tests, including partially funded and uncommitted review budgets. Full verification and the 37-test desktop/mobile/Pages browser matrix passed on the preceding rebased revision. Fresh GitHub ingestion completed with 117 leaders and 10,375 accepted events. Exact current-head hosted checks are pending.

This remains a draft until the coordinated carry/index repair lands: the allocation editor must consume validated carriedMinor before final verification. Production deployment evidence is N/A because this PR has not merged or deployed. Local screenshots, keyboard checks, accessibility results, zoom-equivalent reflow checks, and logs are retained outside Git in /tmp/slop-review-evidence.

Issue #368, opened 2026-09-04
Auto-merge funding records the deterministic verifier already proves


drew-dot-com:
## Observed

`funding/README.md` defines `self-reported` and `verified-on-chain` as separate states and ships read-only verifiers for Solana, Base, Ethereum, and Bitcoin that fail closed. A funding record still enters the ledger only by a pull request that a human merges. Today that human is one of two people, and the merge adds no information the verifier did not already produce.

## Proposal

Run the applicable verifier in CI on any pull request that only appends records under `funding/<project>/**` and, when every added record's verifier output matches the record's recipient, asset, amount, finality, and transaction identity exactly, allow the unattended approver already proven on #302 to approve and merge it. Any other change in the same PR, any `self-reported` record, any `disputed` record, any `supersedes` chain that does not resolve, and any verifier failure leaves the PR on human review.

Self-reported records stay human-merged on purpose. They carry no chain evidence and the README already forbids summing them with verified totals.

## Boundaries

- The verifier remains read-only. It never signs, broadcasts, or writes a record.
- The approver merges only PRs whose diff is confined to the funding path and whose records are all `verified-on-chain`.
- Commitment records (`funding/<project>/commitments/**`) are excluded from this issue until #333 settles the instrument rules.

## Acceptance

- A PR adding one verified Solana USDC record merges with no human action.
- The same PR with the amount off by one minor unit stays open with a failing check.
- A PR touching a manifest and a funding record together stays on human review.
- The approver's decision log names the verifier version and the exact output hash it matched.

Parent: #114. Related: #302.

lalalune:
I am taking the trusted-base funding eligibility and verifier-evidence gate in an isolated branch. Scope: append-only direct-funding records, exact-head evidence, no commitment records and no payment signing or broadcasts. I will keep unattended approval disabled unless its separate protected approver can verify this gate and all required checks. This coordinates the lane with the other active PR work.

lalalune:
Final queue review: #380 and #386 now provide the trusted-base read-only gate, accept increasing confirmations without weakening finality, and reject transaction reuse across direct-funding and commitment ledgers. Exact integrated head passed 873 tests, 125 skill checks, and 39 browser checks. This issue remains open because the requested durable unattended merge authority is not implemented by a read-only decision or by this one-off user-authorized review. The existing external deployment approver is not a funding-PR merger. No real funding-record canary or permanent approver configuration is claimed.

lalalune:
[...]

PR #386, opened 2026-09-05
fix: integrate reviewed funding, wallet safety, persistence, and UI fixes

### LINKED ISSUE #331: Pledged pools display the cap as if funded

### LINKED ISSUE #353: Receipt CLI cannot recover a run after a transient private-trace upload failure

### LINKED ISSUE #366: Compromised-wallet hold: withhold an unsafe payment without forfeiting the amount or restarting review

lalalune:
## Final reviewed integration

Combines #375, #383, #382, #385, and the reviewed UI-only revision of #370 on merged #384. This retains the individual commits and integrates their shared validators before final testing.

Merge-specific work preserves both canonical Solana signatures and settlement chronology, carries unsafe-wallet history through unfunded/legacy months without inventing monetary balances, and updates hold test fixtures to declare synthetic reviewed funding. Pledged production pools remain unfunded. UI clipboard, exact-month allocation limits, separate reward-line limits, and hover contrast fixes are included.

Component evidence: funding full verification 782 Vitest + 124 skill tests; persistence 782 + 125 and independent real SQLite rollback/race validation; hold 797 + 124 before funding integration; monotonic confirmation gate 50 focused tests and full verification; UI 37 viewport + 2 Pages browser tests with 3 intentional duplicate reflow skips, plus 11 funded-fixture cases.

Exact final head a7ce492be98416fa27a029fe9857503da2cfdc0c passed the complete `bun run verify`: 77 Vitest files, 873 tests, 125 skill checks, static checks, and build. The combined hold suite also passed all 64 focused tests; the final gate passed 61 focused tests. Independent review and real SQLite rollback/race checks passed. Exact-head browser checks passed in an isolated worktree: 37 viewport tests and 2 Pages/artifact tests, with 3 intentional duplicate reflow skips. An additional 11 funded-UI cases passed; the final candidate preserves all 92 browser-relevant source/config/test hashes. Actions are explicitly waived by the maintainer. No deployment, live provider identity availability, wallet signing, or settlement is claimed.

Closes #331.
Closes #353.
Closes #366 through the signed-report, immutable-wallet, shared-principal carry implementation and regression coverage. This does not claim a production deployment or apply a hold to any real contributor.
Addresses one correctness defect under #368, not durable unattended merge authority.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -143,6 +143,8 @@ jobs:
           bun scripts/prepare-private-intake-attestation.ts --ensure-disabled
 
       - name: Validate reward lifecycle
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: bun run cycles:check
 
       - name: Validate finalized Solana evidence on trusted revisions
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -143,8 +143,6 @@ jobs:
           bun scripts/prepare-private-intake-attestation.ts --ensure-disabled
 
       - name: Validate reward lifecycle
-        env:
-          GH_TOKEN: ${{ github.token }}
         run: bun run cycles:check
 
       - name: Validate finalized Solana evidence on trusted revisions
--- a/.github/workflows/monthly-rewards.yml
+++ b/.github/workflows/monthly-rewards.yml
@@ -109,8 +109,6 @@ jobs:
         run: bun run rewards:close-month -- --cycle "${{ steps.cycle.outputs.cycle }}"
 
       - name: Verify immutable cycle chain
-        env:
-          GH_TOKEN: ${{ github.token }}
         run: bun run cycles:check
 
       - name: Open review PR
--- a/.github/workflows/monthly-rewards.yml
+++ b/.github/workflows/monthly-rewards.yml
@@ -109,6 +109,8 @@ jobs:
         run: bun run rewards:close-month -- --cycle "${{ steps.cycle.outputs.cycle }}"
 
       - name: Verify immutable cycle chain
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: bun run cycles:check
 
       - name: Open review PR
--- a/.github/workflows/unsafe-destination-transitions.yml
+++ b/.github/workflows/unsafe-destination-transitions.yml
@@ -1,0 +1,48 @@
+name: Trusted unsafe destination transition gate
+
+on:
+  pull_request_target:
+    branches: [develop]
+    types: [opened, synchronize, reopened, ready_for_review]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: trusted-unsafe-destination-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  unsafe-destination-transitions:
+    name: Trusted unsafe destination transition gate
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    steps:
+      - name: Checkout immutable trusted base only
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          fetch-depth: 1
+          persist-credentials: false
+          submodules: false
+
+      - name: Install pinned Bun for trusted local modules only
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6
+        with:
+          bun-version: "1.3.14"
+
+      # Only Git objects are read from the proposal. Never checkout its tree,
+      # install its dependencies, or execute any contributor-controlled code.
+      - name: Preserve accepted cycle and unsafe destination evidence
+        env:
+          CYCLE_BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          CYCLE_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          CYCLE_PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          set -euo pipefail
+          test "$(git rev-parse HEAD)" = "$CYCLE_BASE_SHA"
+          git fetch --no-tags --depth=1 origin \
+            "+refs/pull/$CYCLE_PR_NUMBER/head:refs/remotes/origin/slop-cycle-head"
+          test "$(git rev-parse refs/remotes/origin/slop-cycle-head)" = "$CYCLE_HEAD_SHA"
+          bun --no-install scripts/check-unsafe-destination-transitions.ts \
+            "$CYCLE_BASE_SHA" "$CYCLE_HEAD_SHA"
--- a/.github/workflows/unsafe-destination-transitions.yml
+++ b/.github/workflows/unsafe-destination-transitions.yml
@@ -35,6 +35,7 @@ jobs:
       # install its dependencies, or execute any contributor-controlled code.
       - name: Preserve accepted cycle and unsafe destination evidence
         env:
+          GH_TOKEN: ${{ github.token }}
           CYCLE_BASE_SHA: ${{ github.event.pull_request.base.sha }}
           CYCLE_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
           CYCLE_PR_NUMBER: ${{ github.event.pull_request.number }}
--- a/backend/trace/cloudflare-persistence.ts
+++ b/backend/trace/cloudflare-persistence.ts
@@ -735,6 +735,7 @@ export class CloudflareTracePersistence implements TracePersistence {
 
   async createWalletClaim(
     claim: WalletClaim,
+    audit: AuditInput,
   ): Promise<PersistenceResult<WalletClaim>> {
     const byDigest = await this.db
       .prepare("SELECT * FROM wallet_claims WHERE record_sha256 = ?")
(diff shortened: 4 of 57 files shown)


lalalune:
Exact final head a7ce492be98416fa27a029fe9857503da2cfdc0c completed bun run verify successfully under pinned Node 24.15.0: 77 Vitest files / 873 tests, 125 skill tests, all project/evaluation/funding/cycle/protocol/dependency/type/format/lint checks, and build. Worktree is clean. Independent review confirmed integrated funding/hold history, transaction chronology, primary-signature checks, persistence rollback/race semantics, and the final cross-ledger gate fix. Exact-head browser artifact validation is finishing in a separate detached worktree to avoid shared-build output races. Actions are waived, not claimed green; no deployment is claimed.

lalalune:
Merge verified: develop is 81a3d652cd6e7ab3ad9a6e43c9636f518c3d7a19. Its Git tree bfb3b67aa445558684c8797cb1adba8748742eff exactly equals tested head a7ce492be98416fa27a029fe9857503da2cfdc0c; tested-head ancestry is confirmed. Exact-head canonical E2E completed exit 0 (37 viewport + 2 Pages/artifact tests, 3 intentional duplicate reflow skips). Full verify completed exit 0 (873 tests + 125 skill checks). Issues #331, #353, and #366 closed through this implementation. Production deployment and independent authority/operational issues remain separate.

PR #399, opened 2026-09-06
fix: bind settlement sources to frozen funding instruments

lalalune:
Refs #333. The shared settlement constructor and stored-plan validator previously accepted a caller-chosen Solana source independently of the frozen funding instrument. Require the exact frozen Squads vault, and refuse an unfunded or Sablier/EVM basis as authority for an arbitrary Solana source. Preserve historical allocations predating the funding-basis schema; no committed cycle artifacts are rewritten. Three new regressions fail before the fix and pass after it; the focused plan/exact-cycle suite passes 11 tests. Source tampering is rejected during both creation and readback. No signing, transaction retirement, carry, or payment activation is added. Full exact-head verification and hosted browser checks pending; draft until evidence is complete.

### DIFF
--- a/cycles/README.md
+++ b/cycles/README.md
@@ -208,3 +208,12 @@ bun run cycles:verify
 
 No command reads a private key or signs a transaction. Keep seed phrases and
 private keys out of Git, issues, CI, skills, prompts, and local telemetry.
+
+For an allocation with a frozen funding basis, a Solana execution plan must
+use that exact Squads vault as `sourceOwner`. The same check applies when
+reading a stored plan; another valid wallet is not a substitute. A pledged or
+Sablier/EVM basis cannot produce a Solana plan without a separately reviewed
+funding transition. Historical allocations predating the funding-basis schema
+retain their existing validation; this does not migrate or rewrite records.
+Source binding does not prove current backing, signing capability, transaction
+retirement, or safe carry, and does not enable payments.
--- a/protocol/signer-access-attestations.md
+++ b/protocol/signer-access-attestations.md
@@ -103,9 +103,12 @@ remaining principal and prevent both an old intent and a successor cycle from
 paying it. Only eligible shared-pool principal carries, exactly once; review
 budget does not carry. The successor monthly cycle requires a fresh instrument.
 
-The current unsigned execution-plan schema does not bind `sourceOwner` to the
-frozen funding instrument or record a signed transaction's lifetime, nonce, or
-Squads proposal retirement. Consequently an issued or partially settled plan
+New funding-basis Solana plans bind `sourceOwner` to the exact frozen Squads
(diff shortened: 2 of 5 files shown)


lalalune:
Read-only source review of exact head b765c6310c8e79b14066310f060b94bd602142b3: no blocking findings in the four-file scope. Confirmed assertRewardAllocationManifest validates the full instrument identity before constructor extraction; component 4 is the frozen vault, not the multisig or a member. The shared constructor is reused by assertSettlementExecutionPlan, which is called by cycle validation and settlement verification, so the guard also rejects hand-edited sourceOwner JSON. New tests cover mismatched creation/readback and both EVM networks; existing no-basis tests preserve legacy behavior. Full hosted/local verification remains the implementation lane responsibility and is still pending. This is not evidence of transaction retirement, carry safety, funding availability or payment activation.

lalalune:
The first full hosted run found one integration fixture that used a synthetic Base/Sablier funding basis to create a Solana transfer plan. Updated only that fixture to a synthetic Squads vault matching its existing source wallet, retaining all original-wallet, other-row, review-clock and held-payment-exclusion assertions. The source guard is unchanged. Focused constructor/readback plus unsafe-destination integration now passes 16 tests. Full verification is being rerun on the new exact head; the failed previous run is retained as evidence, not treated as a pass.

lalalune:
Reviewed final head b11456715ce5d5f19f74496759112bc9627463b8 against previously reviewed b765c631. The follow-up changes only the unsafe-destination integration fixture from a Base/Sablier basis to a synthetic Squads vault matching its existing Solana source. The production source-binding guard and all hold, wallet, unaffected-row and review-clock assertions are unchanged. No blocking source-review finding. Full exact-head validation is still required; this comment does not claim it passed.

lalalune:
Final exact-head acceptance: b11456715ce5d5f19f74496759112bc9627463b8 remains based on current develop 0bd8d2874beb3b32195ab15e05d7155ae8c7a32c. Local bun run verify completed successfully: registry/protocol/dependency/type/format/lint checks, 983 unit tests across 85 files, 129 skill tests, and a 175-file production build. Hosted run https://github.com/SlopDotCash/slopdotcash/actions/runs/34021951902 is terminal success with the same 983/129 counts plus 41 viewport tests and 2 Pages/artifact tests. All trusted gates are green; no unresolved review threads. Independent final-head source review found no blocker. The three new regressions failed before the fix; the corrected Solana integration fixture preserves every original hold/wallet/clock assertion. UI redesign review is N/A: no UI implementation changes; the normal browser suite passed. Production deployment and deployed-byte readback remain post-merge acceptance, not claimed here. #333 remains open for transaction retirement and safe hold/carry; no funds or payment authority were created.

PR #415, opened 2026-09-08
fix(ingestion): narrow the GraphQL page when GitHub cannot build it

### LINKED ISSUE #414: Bug: scheduled ingestion fails on every run when GitHub 502s building a 100-node search page
## What happened

Every scheduled `slop.cash` run since `2026-09-06T20:29Z` has failed in
`leaderboard:generate`. That is eight consecutive scheduled failures, most
recently run 34279451740 at `2026-09-08T21:14Z`. The published snapshot is
frozen at `generatedAt 2026-09-06T19:02:58.135Z`.

The cause is not a transient edge fault. `LeaderboardSearchReferences` asks
GitHub Search for a 100-node page of

```text
repo:SlopDotCash/proximityprize is:pr is:merged merged:<35-day rolling window>
```

GitHub times out building that page and nginx returns a `text/html` 502 after
about 10.5 seconds. `execute()` classifies 502 as transient and repeats the
identical request four more times, each failing the same way, then fails
closed. Roughly eight minutes of already-completed hydration is discarded on
every run.

## Expected behavior

A page GitHub cannot build inside its budget should be requested in smaller
pieces rather than repeated unchanged, and the scheduled refresh should
publish.

## Reproduction

Unmodified `develop` at `bdbceec`, run locally against live GitHub from a
different network and a different token than Actions, reproduces the identical
failure at `t+239s`.

Probing that exact query at varying page sizes is the whole finding:

| pageSize | result | elapsed |
| --- | --- | --- |
| 100 | 502 `text/html` | 10688 ms |
| 50 | 200, `issueCount` 98 | 4122 ms |
| 25 | 200, `issueCount` 98 | 2541 ms |
| 10 | 200, `issueCount` 98 | 751 ms |
| 1 | 200, `issueCount` 98 | 654 ms |

The result set is 98 items, so it is a single page at `GRAPHQL_PAGE_SIZE`, and
that is precisely the page GitHub cannot build. Cost scales with the size of
the PR bodies pulled into one page, so this is a threshold the repository
crossed rather than a state that can be waited out.

## Evidence

Instrumented request trace from the unmodified generator, wrapping
`globalThis.fetch` only to observe:

```text
[req 126] LeaderboardSearchReferences -> 502 text/html in 10539ms pageSize=100
[req 127] LeaderboardSearchReferences -> 502 text/html in 10705ms pageSize=100
[req 128] LeaderboardSearchReferences -> 502 text/html in 10736ms pageSize=100
[req 129] LeaderboardSearchReferences -> 502 text/html in 10428ms pageSize=100
[req 130] LeaderboardSearchReferences -> 502 text/html in 10942ms pageSize=100
```

Five identical attempts, about 52 seconds, then the snapshot fails.

Actions logs show the same terminal error on every scheduled run since
`2026-09-06T20:29Z`:

```text
SyntaxError: JSON Parse error: Unrecognized token '<'
  at execute (scripts/generate-leaderboard.ts:1480)
error: GitHub GraphQL HTTP 502 returned a non-JSON response (text/html)
  at execute (scripts/generate-leaderboard.ts:1493)
```

Failing run IDs: 34279451740, 34252735748, 34220740084, 34188245369,
34163644138, 34123483864, 34084658055, 34058139957.

Two diagnosability gaps made this expensive to locate. The generator emits
nothing between start and its single completion line, so an eight-minute run
dies silently. The terminal error reports neither the attempt count nor which
query or repository failed, so a 502 that exhausted retries is indistinguishable
from one that was never retried. Finding it required instrumenting `fetch` and a
full re-run.

## Environment

Bun 1.3.14, Node 24.15.0, ubuntu-24.04, in Actions and locally.

## Relationship to #302

This is distinct from both causes documented in #302. That issue's current
stated next step is provisioning `CLOUDFLARE_API_TOKEN` and
`CLOUDFLARE_ACCOUNT_ID` in the `slop-data-refresh` environment. Scheduled runs
no longer reach that step: since `2026-09-06T20:29Z` they fail inside the
`quality` job at `leaderboard:generate`, before any deploy job starts. So
provisioning those secrets alone will not produce a scheduled publication until
this is fixed.

wbaxterh:
Closes #414.

## The change

On a 502/503/504, halve `pageSize` before backing off instead of repeating the
identical request. The opaque `after` cursor stays valid because a smaller
`first` returns fewer nodes from the same position, so callers page more times
over an unchanged result set. `issueCount` and `totalCount` are independent of
page size, so the existing coherence checks in `collectSearchReferences`
(`page.totalCount !== expectedCount` and the final
`deduped.length !== expectedCount`) still hold. Requests that carry no
`pageSize` retry unchanged, and exhausting the attempts still fails closed.

Second, smaller change: a terminal non-JSON response now names the attempt
count and the final page size. The old message said only
`HTTP 502 returned a non-JSON response (text/html)`, which is why this cost a
full instrumented re-run to locate rather than a log read.

## Why not just lower GRAPHQL_PAGE_SIZE

A fixed smaller constant would not have fixed it. In the verifying run, page 1
of that query succeeded at 50 while page 2 still returned 502 at 50 and needed
25, then the remainder succeeded at 100. Per-page cost tracks the size of the
PR bodies that land in each page, so the usable ceiling moves within a single
query and has to adapt.

## Evidence

A complete `leaderboard:generate` against live GitHub at this branch head, run
unattended, now finishes:

```text
[slop.cash] wrote public/data/leaderboard.json
(118 leaders, 9453 score events, 601 GraphQL requests, 3928/5000 points remaining)
```

The narrowing sequence in that run, from the same `fetch` instrumentation used
to diagnose the bug:

```text
[req 126] pageSize 100 -> 502 in 10461ms
[req 127] pageSize  50 -> 200 in  6943ms   50 nodes, cursor -> 50
[req 128] pageSize 100 -> 502 in 10775ms   next page
[req 129] pageSize  50 -> 502 in 10392ms   a fixed 50 would have failed here
[req 130] pageSize  25 -> 200 in  7751ms   25 nodes, cursor -> 75
[req 131] pageSize 100 -> 200 in  5693ms   final 23 nodes
```

Tests: three new cases in `scripts/generate-leaderboard.test.ts` covering the
reproduced boundary (narrows until the page builds), the fail-closed report
(exhausted attempt count and final page size), and the guard that an
unpaginated query retries with its variables untouched.

Gate at head:

- `typecheck`, `format:check`, `lint:check`: clean, no warnings in touched files
- `test`: 996 vitest across 86 files, plus 129 skill-tests across 7 files, all passing
- `build`: succeeds, 1844 modules
- `test:e2e`: `N/A - ingestion script change with no UI surface`
- `audit:dependencies`: fails identically on clean `develop` with my changes
  stashed (vitest GHSA-82fw-gwwq-j7x9, sharp GHSA-rgj7-g3m4-5g8c), so
  `bun run verify` cannot currently pass end to end on any branch. Untouched by
  this change, and worth its own issue if it is not already known.

Note that PR CI will not exercise this fix: the `Generate live contribution
data` step is gated on `github.event_name != 'pull_request'`, so a PR loads the
deployed ledger instead of ingesting. The live run above is the evidence.

## Known limits, deliberately not addressed

- Page size resets to `GRAPHQL_PAGE_SIZE` on each `execute()` call, so this
  repository pays one wasted ~10s 502 per page. Making the reduction sticky per
  document would remove that at the cost of client state. Happy to add it if
  you want it, but it seemed like scope beyond the demonstrated boundary.
- Detail queries batch by `ids` rather than a cursor page, so they cannot
  narrow this way. No such failure is observed today.

Contribution skill revision: N/A - direct platform pull request, not a skill-mediated contribution
— [claude-code-wbaxterh]

### DIFF
--- a/scripts/generate-leaderboard.test.ts
+++ b/scripts/generate-leaderboard.test.ts
@@ -26,6 +26,7 @@ import {
   estimateFirstPageDetailCost,
   estimateReviewFirstPageDetailCost,
   GitHubGraphqlClient,
+  GRAPHQL_PAGE_SIZE,
   type GraphqlExecutor,
   generateLeaderboardFromGitHub,
   LEADERBOARD_QUERY_DOCUMENTS,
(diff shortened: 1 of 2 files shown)


hermesagent270-commits (approved):
Reviewed exact head 1b3e44f7f97444d93c2b7c5688f2b2fae2b31302. The retry narrows only root cursor-paginated reads; existing total-count, identity, cursor, and final deduplication checks remain fail-closed. Focused generator suite passes 68/68. Combined with dependency PR #416, the full gate passes: 996 Vitest tests, 129 skill tests, audit, typecheck, format/lint, validators, and build. I also reproduced a complete live generation at this head with the #416 dependency graph: 118 leaders, 9,449 score events, 604 GraphQL requests, 4,239/5,000 points remaining. No blocking findings. The current red quality check is the dependency audit fixed by #416; rebase/rerun after that lands.

Issue #429, opened 2026-09-10
Scheduled data refresh stalls behind a merged code release awaiting approval, so the intake attestation cannot renew


drew-dot-com:
## Summary

A merged code release that has not yet been approved at `eliza-army-production` holds the `slop-production` concurrency slot. Every scheduled data refresh behind it queues in `pending`, so the leaderboard stops updating and the private-intake attestation stops renewing. If the release stays parked for 49 hours, `private-request-intake` fails closed.

## Observed today (10 Sep 2026)

- 19:54Z: #428 merged to `develop` (65c9999). Push run 34523151254 passed `quality` and is waiting for approval at `eliza-army-production`.
- 20:56Z: scheduled run 34529309768 passed `quality`, then its `deploy` job went `pending` on the `slop-production` group behind the waiting release. It has not published.
- Live attestation `verifiedAt` is 17:57Z (revision 614d983, the #425 release). Intake fails closed at about 12 Sep 19:00Z if nothing publishes before then.
- Each later 6-hourly schedule will queue the same way. Because the deploy job sets `cancel-in-progress: false`, nothing cancels the parked release, which is the intended behaviour per the comment in `deploy.yml`.

## Why a separate concurrency group is not enough

I first assumed giving schedules their own group would fix it. It would not. The schedule-only step `Require already released source and data-only artifact changes` requires the checked-out SHA to already have a successful push or dispatch release. A schedule that runs on an unreleased `develop` head would fail that check rather than publish. So the schedule is blocked twice: once by the slot, and again by the release baseline rule. The second rule is correct, so the fix has to respect it.

## Proposal

When the `develop` head has no successful production release yet, the scheduled refresh should build from the newest `develop` revision that does have one, rather than from head. That keeps every invariant already in the workflow:

- only approved code ever reaches Pages (the schedule would publish data on top of an already approved revision, exactly as it does today for a quiet branch);
- the code release awaiting approval is untouched and still cannot be cancelled by a schedule;
- `Require live release-input equivalence` still runs against the revision actually being published.

Sketch: at the top of the schedule path, resolve `approved_sha` with the same `gh run list --status success` query the baseline step already uses, walking `git log origin/develop` newest-first until a hit, and check that out for the refresh. If no revision on the current branch has ever been released, fail as today.

A smaller alternative is to have the queued schedule fail fast instead of sitting in `pending`, so the stall is visible in run history. That does not renew the attestation, so I would treat it as a fallback rather than the fix.

[...]

PR #440, opened 2026-09-11
fix(intake): fall through to the live GitHub check when the bundle attestation is stale

drew-dot-com:
## TL;DR

Contributor intake currently goes offline whenever the deployed attestation is older than 49 hours, which happens whenever no deploy has published for two days. This PR makes a stale attestation fall through to the live GitHub check that the API already performs when no bundle is present, instead of failing closed on the age of a deploy. Nothing manual is needed to keep intake up any more. Follows the conversation with Drew on 11 Sep about eliminating the manual dependency entirely.

## What changes

- `functions/api/v1/[[path]].ts`: in `privateIntakeStatus`, a missing, malformed or expired bundle attestation no longer returns `unavailable`. It continues to the existing edge-cached, bounded, redirect-rejecting GitHub check. A fresh attestation is still served directly with no external request.
- `src/lib/private-intake-attestation.ts`: comment on the 49 hour constant updated to describe what the window now means. The value is unchanged.
- `functions/api/v1/trace-api.test.ts`: three new tests. Expired attestation plus GitHub enabled returns 200 and hits exactly the GitHub endpoint. Missing attestation plus GitHub enabled returns 200. Expired attestation plus a rate limited GitHub still returns 503 `private_intake_unavailable`.

## What does not change

- The 49 hour window, the attestation format, the build-time verifier and the deploy workflow are untouched.
- Fail-closed on a malformed, redirected or rate limited GitHub response is unchanged and still covered by the existing tests.
- The reviewed bundle attestation remains the fast path. The live check only runs when the bundle cannot vouch.

## Why this is the right layer

The attestation only asserts that GitHub private vulnerability reporting was enabled at the released revision. When it expires, the honest question is "is it enabled now", and the API already knows how to ask GitHub that with a 5 minute edge cache. Refusing contributors because a deploy did not happen answers a different question. Issue #429 describes how a merged release awaiting approval can starve every scheduled refresh; with this change that starvation delays data, not intake.

## Verification

- `bun x vitest run`: 114 files, 1412 tests passed on develop head 85e1d9e plus this change.
- The two fall-through tests fail without the source change and pass with it.
- `bun run typecheck` clean. Biome reports nothing on the three changed files.

Happy for you to take this over and land it under your own authorship if you prefer, since it is release-path code.

### DIFF
--- a/functions/api/v1/[[path]].ts
+++ b/functions/api/v1/[[path]].ts
@@ -115,6 +115,13 @@ async function privateIntakeStatus(
   fetchImpl: typeof fetch,
   now: () => Date,
 ): Promise<PrivateIntakeStatus> {
+  // The reviewed Pages bundle carries a build-time attestation that GitHub
+  // private vulnerability reporting was enabled for the released revision. A
+  // fresh attestation answers without any external request. A missing,
+  // malformed or expired attestation is not itself evidence that intake is
+  // disabled, so it falls through to the bounded live GitHub check below
+  // instead of failing closed on the age of a deploy. Only GitHub's own
+  // answer, or the inability to obtain one, decides availability.
   if (assets !== undefined) {
     try {
       const url = new URL(PRIVATE_INTAKE_ATTESTATION_PATH, requestUrl);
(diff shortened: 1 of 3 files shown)


drew-dot-com:
Superseded by #445, which moved intake renewal onto the identity Worker's existing hourly cron and is proven live as of 2026-09-12T20:17Z. The bundle attestation this PR patched no longer exists after #439, so the branch conflicts with develop. Closing.

PR #445, opened 2026-09-12
fix(intake): renew the private intake observation from the identity Worker cron

drew-dot-com:
## TL;DR

The hourly renewal job from #439 has never succeeded in production. It runs a D1 write under `slop-data-refresh`, whose Cloudflare token is Pages-only, so every run since 12 Sep 00:39Z has failed at `wrangler d1 execute` with an authentication error. The observation is currently written only by code releases, which means optional trace intake fails closed 24 hours after the last release (next: ~13 Sep 10:44Z). This PR moves renewal into the `slop-identity` Worker's existing hourly cron, where the write goes through the Worker's own D1 binding and needs no GitHub-held Cloudflare credential at all. Follows the conversation with Shaw on 11 Sep about eliminating the manual dependency entirely.

## What changes

- `workers/identity/index.ts`: `renewPrivateIntakeStatus` reads GitHub's public private-vulnerability-reporting status (no token needed for a public repo) and upserts `private_intake_status` with the same "never move `verified_at` backwards" guard the script uses. `runScheduledMaintenance` runs identity cleanup and renewal independently: either failing is logged as a fixed string (error-policy:J2) and fails the invocation, without stopping the other. An unreachable, rate-limited, redirected, oversized, or malformed GitHub answer leaves the previous observation untouched so it expires on schedule instead of being refreshed on bad evidence.
- `.github/workflows/private-intake-watch.yml`: becomes a read-only watch. It fails when the public endpoint is not 200, `enabled` is not true, or `verifiedAt` is older than three hours (two missed Worker renewals). No environment, no secrets, mutates nothing.
- `scripts/release-flow.test.mjs`: asserts the watch job has no environment and no Cloudflare token.
- `backend/trace/PRIVATE_INTAKE_RECOVERY.md`, `workers/identity/README.md`: describe the new renewal path and how to diagnose it from Workers Logs.

## What does not change

- The 24-hour acceptance window in `privateIntakeStatus` and the D1 observation design from #439.
- The post-release `renew-private-intake.mjs` step in `deploy.yml`, which still gives an immediate renewal on every code release under the release credentials.
- The Worker's trigger list. Renewal rides on the existing `17 * * * *` cleanup cron, so `verify-identity-schedules.mjs` still sees exactly one canonical schedule and no provisioning step is needed. Adding a second trigger would have failed the release readback, which rejects any non-empty drift by design.

## Validation

- `bun run typecheck`, `bun run lint:check`, `bun run format:check`: clean.
- `bun x vitest run`: 118 files, 1426 tests passed, including 13 new tests for the renewal and scheduled maintenance paths.
- Evidence the current job fails: runs 34662385078, 34674745165, 34686267689, 34695644553, 34705429164, 34712218572 (all `failure`, all "You may have incorrect permissions on your API token" at the D1 execute).
- After this deploys, the first cron at :17 past the hour should move `verifiedAt` on `GET https://api.slop.cash/api/v1/private-request-intake` without any release. I will confirm and post the readback here.

### DIFF
--- a/.github/workflows/private-intake-watch.yml
+++ b/.github/workflows/private-intake-watch.yml
@@ -1,4 +1,4 @@
-name: Private intake health renewal
+name: Private intake health watch
 on:
   schedule:
     - cron: "47 * * * *"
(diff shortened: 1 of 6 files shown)


drew-dot-com:
Deployed as 69b04f8 (release run 34714275494, finished 2026-09-12T20:07:03Z). The release's own renewal stamped `verifiedAt` 2026-09-12T20:06:19Z. The identity Worker cron then moved it to 2026-09-12T20:17:02Z with no release in between (no deploy run started after 19:28Z). `GET https://api.slop.cash/api/v1/private-request-intake` is 200, `enabled: true`, `source: github-public-status`. Intake renewal no longer depends on a release or on any GitHub-held Cloudflare credential. The old renewal job's last failure was 18:46Z; the read-only watch runs next at 20:47Z. This deployment does not activate payments or approve funding, signing, broadcasts, or settlement.

PR #448, opened 2026-09-13
fix(intake): authenticate the Worker's GitHub status read and name skip reasons

drew-dot-com:
## TL;DR

The hourly intake renewal from #445 works, but not reliably. The Worker reads GitHub's private-vulnerability-reporting status anonymously, and Cloudflare Workers share egress addresses across tenants, so GitHub's 60 per hour anonymous budget for the address is regularly used up by others. Workers Logs for `slop-identity` show `slop private intake renewal skipped` at 21:17Z and 22:17Z on 12 Sep, then every hour from 05:17Z through 09:17Z on 13 Sep. `verifiedAt` on `GET https://api.slop.cash/api/v1/private-request-intake` has been stuck at `2026-09-13T04:17:47Z`, and the 24 hour acceptance window would have failed intake closed at ~04:17Z on 14 Sep with no code change anywhere. The same endpoint answers 200 with the anonymous budget nearly untouched from two non-Cloudflare addresses. A conditional request does not help: an exhausted address gets 403 even with a matching `If-None-Match` (verified).

## What changes

- `workers/identity/index.ts`: the read now presents `GITHUB_INTAKE_STATUS_TOKEN` so GitHub bills the request to the token's own 5,000 per hour budget. The token is a fine-grained PAT with the org as resource owner, access to this repository only, and the single permission **Administration: read-only**. That is what GitHub requires from a signed-in caller of this status: anonymous callers and repo admins get the answer, every other signed-in caller gets 403 (verified with a permissionless token, which returned 403 with `x-ratelimit-limit: 5000`). If GitHub refuses the token (401, or 403 that is not a rate limit) the Worker logs the fixed string `slop private intake token rejected` and repeats the read anonymously once, so an expired, revoked, or under-scoped token degrades to today's behaviour instead of stopping renewal. A blank or malformed secret is treated the same way without being sent. The token is never logged.
- Skipped renewals now log `slop private intake renewal skipped: <reason> [status]` with a bounded reason (`unreachable`, `rate-limited`, `unauthorized`, `rejected`, `malformed`), so the logs say why GitHub was not accepted instead of only that it was not. Rate limits are recognised from 429, or 403 with `x-ratelimit-remaining: 0` or `retry-after`.
- `.github/workflows/deploy.yml`: the exact identity Worker secret inventory gains `GITHUB_INTAKE_STATUS_TOKEN`.
- `workers/identity/README.md`, `backend/trace/PRIVATE_INTAKE_RECOVERY.md`: provisioning line for the new secret, what it is for and why it needs that one permission, and the new log lines.

## What does not change

- The 24 hour acceptance window, the D1 observation design from #439, the `17 * * * *` trigger, and the post-release renewal step in `deploy.yml`.
- What the Worker does with GitHub's answer: an unreachable, rate-limited, redirected, oversized, or malformed response still leaves the previous observation untouched.
- The identity Worker's OAuth path. The token is only sent to `api.github.com` for the status read.

## Operator step before merge

The release's secret inventory check is exact, so `GITHUB_INTAKE_STATUS_TOKEN` must exist on the `slop-identity` Worker before this merges, and it must not be added long before, since any other release in between would fail the inventory check on the unexpected entry. Owned by a maintainer or the operator organisation, never a contributor. I will set it right before merging and post the readback here.

## Validation

- `bun run typecheck`: clean. `bun x @biomejs/biome check workers/identity backend/trace`: clean.
- `bun x vitest run`: 118 files, 1437 tests passed, including 11 new or extended tests: token sent as bearer only when configured, anonymous retry on 401 and non-rate-limit 403, no retry when the token itself is rate limited, blank or malformed secret never sent, each skip reason and status, and the token absent from every log line.
- After the release, the cron at :17 past the hour should keep moving `verifiedAt` with no `skipped` lines in Workers Logs. If a `token rejected` line appears, the token needs its permission or rotating, but renewal continues on the anonymous budget.


drew-dot-com:
Readback. Merged as 27c6281, release run 34753059596 succeeded 11:22Z with the exact secret inventory accepting `GITHUB_INTAKE_STATUS_TOKEN`. The deploy's own renewal stamped `verifiedAt` 11:22:48Z. The Worker cron then moved it to 2026-09-13T12:17:09Z with no release in between, and Workers Logs show no `skipped` or `token rejected` lines since the release.

Token in place: fine-grained PAT, resource owner SlopDotCash, this repository only, Administration read-only, nothing else. A permissionless token was verified to get 403 on this endpoint, and a conditional GET from an exhausted address also gets 403, which is why a token rather than an ETag.

PR #451, opened 2026-09-14
feat(site): add "Who builds on Slop" section to /sponsors

drew-dot-com:
Adds the "Who builds on Slop" section to `/sponsors`, the follow-up promised when #449 landed. A would-be funder can now see what kind of code the leaderboard's contributors actually work on before committing a pool.

## What the section shows

Two live facts computed from the published leaderboard snapshot at render time, the same way `/models` does it:

- **Where scored work lands.** Share of scored ledger events per registered repository over the current window, each labelled with the repository description from its own reviewed manifest. Today that reads 82% `elizaOS/eliza` (core elizaOS agent framework and runtime), 18% `elizaOS/asi` (continual-reinforcement-learning research framework in JAX), and under 1% each for `heirlabs/element-sdk` and the Proximity Prize repository. Nothing is hardcoded; a new manifest shows up on its own.
- **How much of the merged work names a model.** Reuses `summarizeModelOutcomes` totals, so the figure is identical to the one on `/models`.
- **Contributors scored in the window.**

One dated fact from outside the site, clearly labelled as such and linked:

- The 8 September 2026 cross-reference of the 119 contributors then on the leaderboard against their merged pull requests elsewhere on GitHub in 2026. AI agents or LLM tooling was the primary focus for 74% of the 86 with classifiable public work (both denominators printed), and 42% of all 119 had merged into an outside AI repository this year. The copy states the method limits (public data only, keyword classifier, two mass pull-request accounts excluded) and says in so many words that this site does not recompute it. Link: https://who-builds-on-slop.vercel.app (aggregate only, no per-person rows).

What is deliberately not on the page: named outside repositories, provider mix (that is `/models`), the top-20 cohort figure (small n, and it includes lalalune), raw merged-PR counts (they include tiny PRs), and anything from the two mass-PR accounts.

## Changes

- `src/lib/who-builds.ts` + test: `summarizeWhoBuilds(snapshot)` counts ledger events per repository, resolves each to its manifest via `findProjectByRepositoryId`, busiest first, deterministic tie-break. Unregistered repositories stay visible with no invented manifest.
- `src/App.tsx`: `WhoBuildsOnSlop` section on `SponsorsPage`, placed above "See who shows up before you commit." `SponsorsPage` now takes the snapshot state, and `sponsors` is removed from the no-snapshot route list. `DataNotice` renders inside the section, so a failed or stale snapshot is visible and the prose still renders.
- `tests/app.test.tsx`: sponsors test now mocks the snapshot; new test asserts the per-repository lines from the fixture, the dated copy, the "not recomputed" caveat, and the external link attributes.
- `tests/e2e/site.spec.ts`: the sponsors route test also asserts the new heading.

## Verification

- `bun run typecheck`, `bun run check:unused`, `bun run lint:check`: clean.
- `vitest run`: 1,452 tests, all passing once the new module was committed (the git-checkout packaging test builds from a clone, so it fails on untracked files by design).
- Playwright `desktop-chromium`, sponsors route test including the new heading: passing.
- Rendered locally against the dev snapshot and checked the section by eye. Live figures on production today: 82% eliza, 18% asi, under 1% each for the other two, 51% of merged pull requests naming a model, 108 contributors in the window.


wbaxterh (approved):
Reviewed at head. `summarizeWhoBuilds` is clean: it derives everything from the published snapshot, invents nothing, resolves labels through the reviewed manifest, omits zero-event repositories, and breaks ties on repository id so the render order is stable. The three unit tests cover the registered, unregistered and empty cases. I checked the counting against the live snapshot and the arithmetic holds.

Two things I would want resolved before this is the page a funder reads, neither a defect in the logic.

**1. The static paragraph and the live line will disagree on screen.** The dynamic summary renders `snapshot.leaders.length`, which on the current published snapshot is **108**. Two paragraphs later the fixed text says "the **119** contributors then on this leaderboard", dated 8 September. Both are true and the prose does say "then", but they sit inches apart in the same section, and the whole point of the section is that a sponsor trusts the numbers. Every percentage in that paragraph is also frozen against the 119 denominator while the line above it moves with each refresh. Either pin the live line to the same dated snapshot, or say the window count and the cross-reference count are from different dates in the sentence itself rather than relying on the tense.

**2. The evidence for the sponsor-facing claim lives off-repo and unpinned.** `WHO_BUILDS_SNAPSHOT_URL` points at `who-builds-on-slop.vercel.app`. It is live and returns 200, but grepping this repository for it returns zero hits, so the dataset behind "74% of the 86 with classifiable public work" exists only on a third-party deployment that can change or disappear with no record here.

That sits awkwardly against the rest of the protocol. Everything else money-adjacent in this repo is hash-pinned and immutable: run receipts, funding preparations, signed traces. Your own #454, open right now, will not accept an external award without an `archiveUrl` and a `contentSha256` precisely so a source cannot be edited after the fact. The sponsors page is the one surface where someone is deciding whether to commit funds, and it is currently the one citing an unpinned URL.

Cheapest fix consistent with #454: commit the snapshot JSON and its method note into the repo, or at minimum record its sha256 next to the link so the page can state what it is pointing at. That also removes the "it is not recomputed by this site" caveat, since a committed artifact is self-evidently a fixed one.

**Minor:** `RepositoryFootprint.share` is computed and exported but `App.tsx` recomputes the same ratio inline through its own `percent` helper, so the field is currently unused. Not worth a change on its own, just flagging that the two could drift if rounding rules ever differ.

Approving, since the module is correct and both points above are editorial and provenance calls rather than bugs.

Contribution skill revision: N/A, review written without the contribution skill installed

drew-dot-com:
@wbaxterh both editorial points addressed in the new commit, thanks.

**1. The two counts.** The paragraph now opens by saying the contributor count above moves with every refresh and that its own figures are frozen to the leaderboard as it stood on 8 September 2026, when it held 119. Every figure in the paragraph is rendered from one constant (`WHO_BUILDS_CROSS_REFERENCE`) rather than hand-typed prose, so the denominators cannot drift from each other.

**2. Provenance.** The aggregate-only snapshot is committed at `data/who-builds/2026-09-08/snapshot.json` with a `METHOD.md` beside it (inputs, method, headline table, caveats). The page links the committed file and shows its SHA-256 (`540df0156b3b...`), plus the method note, with the Vercel page demoted to "Rendered view". A unit test hashes the committed file and checks every quoted figure (119, 86, 64, 50, the two excluded accounts, the 74/54/42 roundings) against it, and the formatter is told to leave `data/who-builds` byte-exact the way it already does for `cycles/`.

**Minor.** The page now renders `RepositoryFootprint.share` directly through one rounding helper, so the module and the page cannot drift.

AI provider/model: anthropic / claude-fable-5-1

PR #454, opened 2026-09-15
feat(evaluations): opt-in awards for contributions outside GitHub

drew-dot-com:
## Summary

Extends the existing `evaluations/` award path so a maintainer can recognize useful work that never touches the repository: a public explainer thread, a support answer that closed a recurring question, a tutorial, or a video.

Nothing changes for any current project. A project opts in with one reward field, absent by default:

```json
"externalEvaluations": { "enabled": true }
```

With the opt-in, an award file may use `source.kind: "external"`:

- `source.platform` is one of `x`, `discord`, `youtube`, `web`, and `source.url` must be the canonical public address for that platform (no query string, no fragment, GitHub URLs rejected here).
- `source.id` is `external-` plus the sha256 of the exact URL, so the same URL cannot be awarded twice under a different id.
- `source.evidence` is required: a `web.archive.org` capture of the exact URL (or an `archive.ph` snapshot), the sha256 of the captured content, and a capture time between `occurredAt` and `review.reviewedAt`. The award stays auditable after deletion.
- No `source.number`, no run receipt, no evidence bonus. Nothing was run against the repository, so there is nothing to sign.

Everything else is the existing award contract: the same one-file PR, the same 1 to 8 points, the same maintainer-only decision, the same newest-three cap per contributor and project, the same canonical-source dedupe. Reach metrics (likes, views, reposts) are explicitly not a reason; the maintainer states the concrete outcome.

Why: contributors keep asking how non-code help counts, and the honest answer today is "it does not, unless it lives in a GitHub comment". This routes that work through the one gate that already exists instead of adding a submission queue, an X API dependency, or a second review process. Submission volume stays bounded because only a merged maintainer PR creates score.

Files:

- `src/lib/external-sources.ts` (new): shared canonical-URL, archive-evidence, and id rules.
- `src/lib/evaluator-awards.ts`: accepts `external` only when the project opted in; validates platform, URL, id binding, and evidence timing.
- `src/lib/leaderboard.ts`, `src/lib/leaderboard-types.ts`: public snapshot validation for the new kind; `platform` and `evidence` are rejected on every other kind.
- `src/lib/project-schema.mjs`, `src/lib/projects.d.mts`: optional `reward.externalEvaluations` with exact keys.
- `evaluations/README.md`, `protocol/scoring-v2.md`: contract and example.

## Validation

```
bun run typecheck                      pass
bun run lint:check                     37 warnings, identical count on develop baseline (none in changed files)
bun run format:check                   1 warning, identical on develop baseline (funding/quality/eliza-2026-08.json)
bun run check:unused                   pass
bun run evaluations:check              validated 11 evaluated contribution awards
bun run projects:check                 project registry is current
bun run cycles:check                   validated 1 reward cycle
bun x vitest run                       1448 passed, 9 skipped (the two temp-copy integration suites pass once the new file is tracked)
bun run build                          built, 184 public files
```

New tests: `src/lib/external-sources.test.ts`, `src/lib/evaluator-awards.external.test.ts`, plus cases in `leaderboard.test.ts` and `project-schema.test.ts` covering opt-in refusal, id binding, GitHub-URL rejection, evidence timing, and smuggled fields on ordinary kinds.

## Evidence

- Before screenshots: N/A - no UI markup changed; the event list already renders `source.url` and `evaluation.decisionUrl`.
- After screenshots: N/A - same reason.
- Walkthrough video: N/A - validator and schema change only.
- Backend logs: N/A - no worker or function touched.
- Frontend console/network logs: N/A - no runtime path changed.
- Real-LLM trajectory: N/A - not a scored contribution.
- Domain artifacts: test output above; no award file added in this PR.

## Contribution provenance

- AI assistance: yes
- AI provider/model: Anthropic / claude-fable-5-1
- Client / agent tooling: Claude Code
- Contribution skill revision: N/A - repository maintainer change
- Attribution status: self-reported

## Slop protocol changes

- Project manifest (`projects/<id>/project.json`): none changed; schema gains optional `reward.externalEvaluations.enabled`.
- Evaluated-contribution award (`evaluations/<id>/award-*.json`): new optional `source.kind: "external"` shape, documented in `evaluations/README.md`.

## Safety review

- [x] No private key, seed phrase, API token, raw private trajectory, or secret was added.
- [x] Untrusted project code is not executed with repository or deployment credentials.
- [x] New telemetry is bounded, disclosed, project-attributed, and covered by adversarial tests.
- [x] Money state is derived from validated manifests and finalized on-chain evidence, not a transaction signature alone.


wbaxterh (changes requested):
Reviewed at head, focused on `src/lib/external-sources.ts`, since that file is the whole new trust surface. The shape is right: opt-in per project, maintainer-merged, no run receipt, no evidence bonus, and one shared validator so review and publication cannot disagree. The canonical-URL idea is the correct primitive.

Two findings below are blocking, because both let the same work be awarded twice. I reproduced all of them against a verbatim transcription of the file (types stripped, every literal, regex and branch copied), so these are observed outputs and not readings.

**1. `web` does not reject GitHub URLs, only `github.com` itself.** The file comment says GitHub URLs must use the ordinary source kinds, and the PR body says GitHub URLs are rejected. `RESERVED_HOSTS` is an exact-host set, so only the apex and `www.` variant are covered:

```
web  https://github.com/elizaOS/eliza/pull/1          -> rejected
web  https://gist.github.com/someone/abcdef123456      -> ACCEPTED
web  https://raw.githubusercontent.com/o/r/main/README -> ACCEPTED
web  https://someone.github.io/writeup                 -> ACCEPTED
web  https://api.github.com/repos/o/r                  -> ACCEPTED
```

A gist is the natural place to put a writeup, so this is not a contrived path. It means GitHub-hosted work can enter through the external lane, which is the one lane with no run receipt and no evidence requirement.

**2. `web` is a general bypass of every per-platform canonical rule, and the ids do not collide.** The `x` branch is strict: `x.com` or `twitter.com`, `/{handle}/status/{id}`, no query. Every front-end mirror of the same post satisfies `web` instead:

```
x    https://x.com/wbaxterh/status/1234567890123456789          -> ACCEPTED
web  https://fxtwitter.com/wbaxterh/status/1234567890123456789  -> ACCEPTED
web  https://vxtwitter.com/wbaxterh/status/1234567890123456789  -> ACCEPTED
web  https://nitter.net/wbaxterh/status/1234567890123456789     -> ACCEPTED
web  https://xcancel.com/wbaxterh/status/1234567890123456789    -> ACCEPTED
```

`externalSourceId` is `sha256(sourceUrl)`, so the canonical URL and each mirror hash to different ids:

```
canonical  external-081dd4f51938055713a2a34b82f2581fc0af3c7fc45dea667e829ec04dc87293
fxtwitter  external-65b9b5b0238ba0af26202bf94dd83168157ca958d7f9b454e7cbc7866b690870
same id?   false
```

The comment on `externalSourceId` says the same URL can never be awarded twice, and that is accurate. The gap is that the unit people get paid for is a piece of work, not a URL string. One X post supports at least five distinct awards today, and with the newest-three-per-contributor-and-project rule that is enough to fill a contributor's entire external allowance from a single post.

Worth noting you already capture the thing that closes this. `evidence.contentSha256` is validated and stored but never consulted for identity. Deduplicating on the content hash as well as the URL id would make the mirror question moot even if the host list is never made exhaustive, which it realistically cannot be.

**3. Non-blocking: the `archive.ph` branch does not bind to the source.** Wayback captures are bound by `capturedPath.endsWith(sourceUrl...)`, and an unrelated wayback capture is correctly rejected. `archive.ph` accepts any snapshot id matching `/^\/[A-Za-z0-9]{5,}$/`:

```
evidence.archiveUrl https://archive.ph/AbCdE
for source          https://x.com/wbaxterh/status/1234567890123456789   -> ACCEPTED
```

So "keeps an external award auditable after deletion" holds for wayback and is advisory for archive.ph. Since archive.ph snapshot ids are opaque and carry no target in the path, I do not think this is fixable in the validator. I would either state the asymmetry in `evaluations/README.md` so a reviewer knows an archive.ph link must be opened and checked by hand, or require wayback whenever the platform has a canonical form.

**4. Minor: some legitimate sources cannot be submitted at all.** The `youtube` branch requires `www.youtube.com` exactly, and `RESERVED_HOSTS` then blocks the `web` fallback:

```
youtube  https://youtube.com/watch?v=dQw4w9WgXcQ    -> rejected, web fallback also blocked
youtube  https://m.youtube.com/watch?v=dQw4w9WgXcQ  -> rejected, web fallback also blocked
```

Both are addresses YouTube itself hands out. Separately, `web` sets `!parsed.search`, so any article URL carrying a query parameter is unsubmittable. Canonicalizing on input would be friendlier than rejecting, and it would also shrink finding 2, since normalizing a mirror to its canonical form makes the duplicate detectable.

Happy to re-review quickly on a push. The first two are the only ones I would hold the merge for.

Contribution skill revision: N/A, review written without the contribution skill installed

drew-dot-com:
@wbaxterh thanks, both blockers reproduced and are fixed in the new commit.

**1. GitHub subdomains.** Reserved hosts are now suffix-matched with every subdomain and the list covers `githubusercontent.com`, `github.io`, `githubassets.com`, the X, Discord and YouTube hosts, and their known front-end mirrors (fxtwitter, vxtwitter, fixupx, fixvx, nitter, xcancel, threadreaderapp, youtu.be, m.youtube.com, youtube-nocookie, discordapp, discord.gg and so on). Your five GitHub cases and the mirror cases are in `external-sources.test.ts` as rejections.

**2. Mirrors and the work unit.** Two layers, since the host list cannot be exhaustive:
- Each platform now accepts exactly one URL form: `https://x.com/<handle>/status/<id>` (twitter.com rejected), `https://www.youtube.com/watch?v=<id>` (youtu.be and m.youtube.com rejected), `https://discord.com/channels/<g>/<c>/<m>`. The rejection message names the accepted form so the maintainer rewrites by hand; the validator never rewrites silently because the stored URL is what the archive evidence and the id are bound to.
- New `externalWorkKey(url, platform)` resolves an accepted URL to the piece of work (X status id regardless of handle, YouTube video id, Discord message ids, exact address for web). `loadEvaluatorAwardEvents` and the snapshot builder both reject a second award on a repository that repeats a work key **or** repeats `evidence.contentSha256`, which is the field you pointed out was stored but never consulted.

**3. archive.ph** is documented in `evaluations/README.md` as unbindable: a reviewer must open it and confirm it shows the source, and Wayback is preferred where the platform allows a capture. I did not make Wayback mandatory for X because archive.org captures of X are unreliable enough that it would close the X lane in practice.

**4.** `web` now accepts a query string where the page needs one. YouTube variants are rejected with the canonical form named rather than silently canonicalized, for the binding reason above.

Full vitest, typecheck, unused and lint are green locally. Re-review welcome whenever you have a minute.

AI provider/model: anthropic / claude-fable-5-1

wbaxterh:
Re-reviewed at `d0ec2400a`. Both blockers are closed. I ran the merged module rather than reading it, transcribing it verbatim and replaying my original cases plus a few new ones.

**Confirmed fixed:**

- All five GitHub cases reject under `web`: `github.com`, `gist.github.com`, `raw.githubusercontent.com`, `someone.github.io`, `api.github.com`.
- All four mirrors from the review reject: fxtwitter, vxtwitter, nitter.net, xcancel.
- `externalWorkKey` collapses the same status id across different handles to one key, so the two-award path is closed at the work level rather than only at the string level. Wiring it into both `loadEvaluatorAwardEvents` and the snapshot builder, and pairing it with `contentSha256`, is a stronger answer than what I suggested.
- Canonical forms still pass, and `web` now accepts a query string, which was the practical complaint in my point 4.

**No false positives from the suffix match**, which was my main worry with `endsWith`. `mygithub.com`, `notgithub.io` and `github.com.evil.example` are all still accepted as `web`, correctly, since none of them is a GitHub host.

**The documented residual behaves as documented.** An unlisted mirror (`nitter.1d4.us`, `twiiit.com`, `xcancel.io`) still enters as `web` and produces a different work key from the canonical X form, so `evidence.contentSha256` is the only thing standing between it and a second award. That matches what you wrote in the module comment, and I agree the host list cannot be made exhaustive. Worth knowing it is load-bearing rather than belt-and-braces.

**One nit, not worth a follow-up on its own.** `resolveWorkKey` keeps `!parsed.search` on the `x` branch but the `discord` branch no longer has it, so `https://discord.com/channels/1/2/3?x=1` is accepted while the X equivalent is refused. Dedupe is unaffected because the discord key ignores the search string, so the only effect is that a stored Discord source URL can carry arbitrary query parameters. Worth folding into whatever touches that file next.

Your call on archive.ph and on not mandating Wayback for X both read as the right trade to me, and documenting the unbindability in `evaluations/README.md` so a reviewer knows to open it by hand is the honest version of that.

Contribution skill revision: N/A, review written without the contribution skill installed

PR #447, opened 2026-09-13
Propose Permission Diff and Visual Strategy Canvas as paused community projects

emanalshazly:
## Summary

Proposes MONNA Agent Permission Diff and Visual Strategy Canvas as two community projects, each with a manifest, a self-contained contributor skill and a separate advisory reviewer skill. Both remain **paused**, with unverified authority, pending receipt activation, unknown copyright/inbound terms, zero funding and disabled payments.

Fixes the compatibility problems exposed by adding projects:

- Historical snapshots retain their collected inventory and immutable reward evidence. Source records and activity must match that inventory; fresh publication still requires the complete current registry.
- PR previews do not fabricate collection for newly registered projects. The UI explicitly identifies uncollected activity, and reward views refuse to calculate against missing source coverage.
- Paused proposals stay unpromoted after collection arrives and no longer claim two unfunded cycles. Long repository names wrap in the agent prompt at narrow widths with enlarged text.
- Registry fixtures derive from manifests; receipt-CLI checks apply to instrumented packages while every skill archive remains validated.

**Merge hold: leave this PR open and unmerged. Auto-merge is disabled.**

## Validation

Head: `15ec5fbf5a13527208c65bbda9b08c63da2e1718`, based on `c5318c017e03c7ad08cda78aaf216dec65485173`. GitHub's merge-ref tree matches the tested head tree, `4cca26af73ed49365ea5b98426907635fe4b4d29`.

- `bun install --frozen-lockfile --ignore-scripts` — PASS, Node 24.15.0 / Bun 1.3.14.
- `bun run verify` — PASS: 119 test files, 1,462 application tests, 115 skill tests, schema/policy checks, dependency audit, typecheck, formatting, lint and build. Lint reports non-fatal warnings.
- `bun run test:e2e` — PASS: 93 preview tests and 3 Cloudflare Pages tests. Three duplicate reflow cases are intentionally skipped because the wide-desktop test exercises all of its own viewport scenarios.
- Desktop/mobile review, keyboard paths, 200% text enlargement, 320 CSS-pixel reflow, WCAG AA checks and raw Markdown/archive contracts passed. Fifty recorded browser diagnostic reports contain no application console or first-party request failures.
- Immutable repository and steward IDs verified against live GitHub. Both commit-pinned MIT license files match their manifest SHA-256 digests. Skill commands and scope checked against upstream READMEs/package scripts.

[...]

PR #457, opened 2026-09-16
fix(registry): collect only active projects into the target repositories

wbaxterh:
Opens the platform half of the question raised reviewing #447, so that proposal is not blocked on platform work. @emanalshazly asked for this to be a separate PR with maintainer agreement on the policy, and their two MONNA manifests stay paused, unfunded and receipt-inactive regardless of what happens here.

## Problem

`TARGET_REPOSITORIES` flattens `PROJECTS` with no filter, so every registered project's repositories enter the ingestion set. The registry records `status`, and `project-schema.mjs` already constrains it to `active` or `paused`, but nothing consumed it at the collection boundary.

It has never mattered because **every registered project is `active` today**. #447 would register the first `paused` ones. As it stands they would be collected on merge, and merges and reviews on them would become scored ledger events on repositories their own steward controls, while the manifest says the project is paused and its authority is unverified.

Money is not affected either way. `assertProjectPaymentsEnabled` independently requires `paymentMode: "enabled"`, `fundingState: "committed"`, `committedMinor > 0` and an exact reviewed fresh-cycle policy. This is about what reaches the ledger and the public leaderboard.

## Change

Extract `collectTargetRepositories(projects)` and filter to `status === "active"` before flattening. `TARGET_REPOSITORIES` is now that function applied to `PROJECTS`.

**Authority is deliberately not part of the gate.** `heir-elements-sdk` is `status: "active"` with `authority.state: "unverified"`, which `project-schema.mjs` explicitly permits while receipts stay pending and payments stay disabled. Gating on authority would silently drop a repository that is collected and scored today. A test pins that.

**`role` stays positional over the collected projects**, so pausing the highest-capped project promotes the next active one to `primary` rather than leaving the registry without one. A test pins that too.

## Compatibility

Every registered project is active, so the published registry is unchanged: the same four rows in the same order, `elizaOS/eliza` still `primary`. The gate is inert until a paused project is registered. That is asserted directly rather than claimed:

```ts
expect(PROJECTS.every((entry) => entry.status === "active")).toBe(true);
expect(collectTargetRepositories(PROJECTS)).toEqual([...TARGET_REPOSITORIES]);
```

## Evidence

**Failure-first.** Deleting only the `.filter(...)` line fails 3 of the 5 new tests, with the paused project's repository appearing in the collected set:

```
× collects no rows for a paused project
    expected [ Array(3) ] to deeply equal [ 'elizaOS/eliza', 'elizaOS/asi' ]
× promotes the next active project when the first one is paused
    expected [ Array(2) ] to have a length of 1 but got 2
× returns nothing when every project is paused
    expected [ { id: 'someone/monna', …(10) } ] to deeply equal []
```

The two that still pass are the compatibility guards, which is the correct behaviour since neither depends on the filter.

**Green at `24a27c1`:**

- `vitest run` full suite: **1464 passed, 0 failed** with `SLOP_PYTHON` pointed at a PyYAML 6.0.3 interpreter. Without it, `scripts/prepare-site.test.ts` and one `check-unsafe-destination-transitions` case fail on the environment; both pass once it is set, so neither is related to this change.
- `bun run typecheck`: clean, no output.
- `bun run projects:check`: `[Slop] project registry is current`.
- `bun run check:unused`: clean.
- `biome check` on the three touched files: clean.

**Disclosed pre-existing red:** `lint:check` and `format:check` do not pass on clean `develop` at `99d0f5a`. I measured both sides rather than assuming: **37 lint warnings and 1 format warning before this change, and 37 and 1 after**. The existing hits are `noDescendingSpecificity` in `src/styles.css` and `noNonNullAssertion` in `src/lib/leaderboard.test.ts`, none of which this PR touches.

## Open question for a maintainer

This implements the reading that a paused project is registered and listed but not collected. The alternative reading is that paused projects should still be collected for rank while excluded from pay. If that is the intent, say so and I will close this and send a `protocol/scoring-v2.md` change documenting it instead, because the manifest fields currently read as gates and a contributor will act on that.

Closes nothing on its own; #447 can proceed once the policy is settled either way.

Contribution skill revision: N/A, authored without the contribution skill installed


drew-dot-com (approved):
Approving. Reviewed at `24a27c1`, ran `repositories.test.ts` and `projects.test.ts` locally (8 passed).

**On the open policy question: paused means registered and listed, not collected.** That is already the reading the README gives. "New projects begin paused. Public contribution access may open independently when missing authority and terms remain explicit, receipts stay pending, and payments stay disabled." Contribution access opening is the transition to `active`. A repository whose merges and reviews land on the public leaderboard has contribution access open by any reasonable definition, so collecting a paused project would contradict that sentence. Your implementation matches the docs; the second reading does not.

Two things I checked specifically:

- Inert against the current registry, and the test pins the four published rows and `elizaOS/eliza` as primary rather than asserting it. Good.
- Gating on `status` alone and not on `authority.state` is right, for the reason you give: `heir-elements-sdk` is collected today and must stay collected.

One follow-up, not blocking this PR: the projects README should say in one line that a `paused` project is listed but not collected, so the manifest field and the code agree in writing. I will take that as a separate commit unless you want it.

@emanalshazly #447 is unblocked by this on the platform side. Your manifests do not need to change.

Issue #455, opened 2026-09-15
August settlement: batch 2 was asi, and both pools silently skipped recipients without a USDC token account (624.72 USDC)


wbaxterh:
## Summary

Two things about the August settlement, both verified on-chain against the immutable preparations.

1. **Batch 2 was the `asi` 2026-08 settlement, not an anomalous eliza batch.** The reconciliation in #406 compared it against `funding/preparations/eliza-2026-08.json`, which is why it read as "17 of 19 wallets, amounts not pro-rata."
2. **Both settlements silently skipped every recipient who had no USDC associated token account.** Six payments never landed, across four contributors, totalling **624.721019 USDC** that the preparations say is owed. Nothing in the public record reflects this.

## 1. Batch 2 is asi

Every amount in the 12 Sep 00:42-00:57 window matches `weight / 17973500 * 5000` against the **asi** preparation, exact to six decimals. 13 of the 15 that landed are exact; the two others are exact plus a small extra transfer in the same window.

The two cleanest proofs are the contributors who appear in the asi preparation and **not** in eliza's at all, so their balance has no eliza component:

| Contributor | asi weight | Entitled | Received | When |
| --- | ---: | ---: | ---: | --- |
| `Prabhat1308` | 1,390,000 | 386.680391 | 386.680391 (+0.0001 probe) | 12 Sep 00:46:12Z |
| `PrototypicNoise` | 10,000 | 2.781873 | 2.781873 | 12 Sep 00:56:49Z |

`Prabhat1308` is the #2 asi contributor and absent from eliza's preparation entirely, so a 386.68 transfer inside batch 2 cannot be an eliza payment.

The reason batch 2 looked like eliza is that 16 of the 18 asi wallet-holders also appear in eliza's preparation, so matching on address alone finds them there.

So the accurate public statement is: **eliza 2026-08 paid 53 transfers / 4,219.71 USDC, and asi 2026-08 paid 15 transfers / 1,348.51 USDC.** `cycles/index.json` still lists only `eliza/2026-07`, so asi's settlement is recorded nowhere.

## 2. Recipients without a USDC token account were skipped

Every wallet that did not already hold USDC received nothing. No failure surfaced anywhere.

**eliza 2026-08** (56 registered wallets, 53 paid):

| Contributor | Weight | Entitled | SPL accounts, any mint |
| --- | ---: | ---: | ---: |
| `hermesagent270-commits` | 3,618,000 | 300.149328 | 0 |
[...]

Issue #406, opened 2026-09-08
Pay out August slop.cash


lalalune:
We need to pay out the september people

lalalune:
Claiming the release lever for PR #425: publish the project funding review, wallet registration route, and vault tracking through the checked-in develop workflow after exact-head checks pass. Eliza's contribution-month caps are August 10,000 USDC and September 5,000 USDC. No DNS, release credential, environment allowlist, or payment activation change is planned. No keys, signing, broadcasts, or payouts are included. This issue remains open until August settlement is verified.

lalalune:
PR #425 is merged and deployed at `614d983be2ae703ea4d5b4240191ce84be5d244a` through protected production run https://github.com/SlopDotCash/slopdotcash/actions/runs/34509712979. Immutable Pages deployment: https://1cc350a9.eliza-computer.pages.dev.

Live review: https://slop.cash/projects/eliza/funding. August shows 10,000 USDC, 108 contributors, 53 missing destinations, and 0 verified paid. September's manifest cap is 5,000 USDC. The app retains all contributors, reasoned exclusion drafts, missing-wallet filtering, registered versus cycle-locked wallet states, and the funding/approval/payment guide. Wallet route: https://slop.cash/wallet.

Validation: 1,378 unit and 129 skill tests; 77 browser cases plus 3 Pages contracts; exact production quality and deployment passed. Independent post-deploy comparison verified all 176 public files and the deployment manifest. Live desktop/mobile funding and wallet reloads on slop.cash returned 200 without overflow or application console errors. Twenty-five read-only API/CORS checks passed. slop.tech wallet loads correctly but its injected Cloudflare analytics beacon is blocked by the strict CSP; this external-console issue remains recorded.

Payment bootstrap preparation is separate draft PR #426, with all hosted checks passed and operator review still required. Branch protection is currently absent; no bootstrap activation, real vault, signer identities, fee recipient, or funded payment test has been completed. A funded personal wallet alone does not satisfy these prerequisites. No payments or real recipient changes were made. Keep this payout issue open until external execution and finalized reconciliation are verified.

GG100-eng:
[...]

Issue #405, opened 2026-09-07
Five additional previously credited reviews are missing after upstream PRs became unavailable


mashingaan:
## Summary

Five additional previously credited Eliza reviews are missing from the published leaderboard after their upstream PR endpoints became unavailable. The accepted-review-history manifest introduced in #320 does not contain these records.

This is a follow-up to closed issue #313, not a claim that its two original restorations failed. I first reported these additional records in [this comment](https://github.com/SlopDotCash/slopdotcash/issues/313#issuecomment-5574337166). This issue provides a dedicated open tracker for their verification and reconciliation.

## Evidence and limits

My locally retained audit of the public snapshot generated at `2026-08-27T23:58:58.242Z` records these accepted `substantive-review` events for GitHub actor `mashingaan` (`U_kgDOCzScNQ`). Each carried 3 score thirds:

| PR in elizaOS/eliza | Review source node | occurredAt (UTC) |
| --- | --- | --- |
| #24021 | `PRR_kwDOMT5cIs8AAAABKhZmaQ` | 2026-08-22T21:22:26Z |
| #25310 | `PRR_kwDOMT5cIs8AAAABKiOa0w` | 2026-08-23T07:59:55Z |
| #25685 | `PRR_kwDOMT5cIs8AAAABKif24A` | 2026-08-23T11:01:01Z |
| #25712 | `PRR_kwDOMT5cIs8AAAABKiksQQ` | 2026-08-23T11:48:32Z |
| #25848 | `PRR_kwDOMT5cIs8AAAABKivXtw` | 2026-08-23T13:28:05Z |

In the [public leaderboard](https://slop.cash/data/leaderboard.json) generated at `2026-09-06T19:02:58.135Z`, these five source IDs are absent. No corresponding exclusion was found in `reviewExclusions`. All five event dates remain inside its August 2-September 6 window.

Authenticated `GET /repos/elizaOS/eliza/pulls/{number}` requests returned HTTP 404 for all five on September 7. This establishes unavailability to the checking account, not who removed an object or why.

Important limitation: the historical baseline here is a locally saved audit of the published snapshot. I have not independently recovered the corresponding immutable Actions artifact for this follow-up. Please verify that authoritative artifact before accepting restoration, as was done for #313.

## Observable impact

Comparing source IDs between those two snapshots:

- Baseline: 21 Eliza review events = 63 thirds.
- Five Eliza review events disappear = -15 thirds.
[...]

PR #458, opened 2026-09-16
feat(funding): reconcile a preparation against observed settled transfers

wbaxterh:
Closes the gap behind #455. A preparation says who is owed what, the chain says what moved, and nothing in the repo compares them.

## Why

The 2026-08 payouts happened outside the settlement pipeline. `verifySettlement` refuses at `assertProjectPaymentsEnabled`, every project is `paymentMode: "disabled"` with `committedMinor: "0"`, and it also requires cycle state `settlement-planned` when no August cycle exists at all. So 72 transfers and 6,004.98 USDC moved with no execution plan, no settlement manifest and no on-chain verification.

That skipped `compileSquadsBatchChild`, which already emits `SQUADS_ATA_PROGRAM` data `[1]`, CreateIdempotent, before each `TransferChecked`. A bare transfer cannot land on an absent associated token account, so three registered asi wallets holding **203.076752 USDC** were skipped silently, plus three more on eliza. Nothing caught it for five days.

## What this adds

`reconcileFundingPayout(preparation, observedByAddress)` classifies every contributor row:

| status | meaning |
| --- | --- |
| `exact` | observed the entitled amount |
| `mismatch` | observed a different non-zero amount |
| `unpaid` | wallet on record, nothing observed |
| `unpayable` | no wallet, so the preparation cannot direct a payment |

Plus a payable subtotal separate from entitlement no wallet can receive, the observed addresses no contributor claims, and a single `reconciles` flag that is true only when every payable row is exact.

Entitlement comes from `createFundingReview`, not from re-deriving `weight / total * cap`. That matters: a failure then means the payout disagrees with the preparation, rather than my arithmetic disagreeing with the repo's. Amounts stay bigint minor units and a non-integer observation is refused instead of coerced.

## Evidence

**It reproduces the real settlement.** Running the allocator against the published `asi-2026-08` preparation and the amounts actually sent on 12 Sep, 13 of 13 paid wallets match to the minor unit, including `Prabhat1308` at `386680391` and `PrototypicNoise` at `2781873`. So the transfers were computed from this preparation by this repo's own allocator, which also settles the open question on #406: that batch was asi, not an anomalous eliza batch.

**Failure-first, two mutations, both caught:**

- `reconciles` ignoring unpaid rows: 2 of 6 fail, including `names every registered wallet that received nothing`.
- an omission scored as `exact` rather than `unpaid`: 1 of 6 fails, on the same test.

**Green at `c538204`:** 6 new tests; full `vitest run` **1479 passed / 124 files, 0 failed** with `SLOP_PYTHON` on a PyYAML 6.0.3 interpreter; `typecheck` clean; `check:unused` clean; `biome check` clean on both files.

**Disclosed pre-existing red:** `lint:check` and `format:check` do not pass on clean `develop`. I measured both sides: 37 lint warnings and 1 format warning before this branch and after, all in `src/styles.css` and `src/lib/leaderboard.test.ts`, none of which this touches.

## One thing this turned up, worth its own look

The same run does **not** reconcile eliza. For the one row I can verify against chain, the allocator produces `7466401` where `7453416` was actually sent, a 12,985 minor gap. asi matches exactly and eliza does not, so the eliza transfers were computed on a basis this repo cannot currently reproduce, possibly a preparation revision that predates the #405/#434 score restorations. One row is not a conclusion, which is rather the point: there is no way to check the other 52 without this.

## Scope

Library only, nothing calls it yet. The natural next step is a thin CLI that fills `observedByAddress` from RPC and fails on any `unpaid` row, and wiring that into the funding workflow, but that is a separate reviewable piece and I did not want to bury the primitive underneath it. Happy to send it next if this shape is right.


drew-dot-com (commented):
Reviewed at `c538204`. Ran the six tests locally, all pass. Then, because this is the first tool in the repo that can answer the question, I ran `reconcileFundingPayout` against the 72 finalized transfers from 11 and 12 Sep, keyed by recipient owner and split by batch. The primitive is exactly the right shape and I want it in. One thing I read differently from #455, which lands in the test fixture, and a few non-blocking notes.

**1. The fixture and the chain disagree, I think.** `WITHOUT_TOKEN_ACCOUNT` lists `2eFE3fee…` (deepanshu-yd) and `73SUereG…` (hermesagent270-commits) as wallets with no USDC account when the asi transfers were made. When I pull the batch-2 transactions, both show a finalized credit to that owner's USDC account:

| Contributor | Amount | Signature | Time |
| --- | ---: | --- | --- |
| deepanshu-yd | 153.003032 | `2yC12hnJfdgJynyz8afX5cpyoEFJVe6YqiDvLxmQ8Jh8PEWmkm8cujvKY1Kvp3mcNXtdpQ3wyPZHKNwtux8QhZSL` | 12 Sep 00:44:01Z |
| hermesagent270-commits | 25.036860 | `4skynN8ot1ithQChmnvTJdE1GR7wY71PGuzi18ZPzyN9SC4CVQGsE79FSjtL8Qcqz7VguM868HYukGE3vXfbcZdB` | 12 Sep 00:55:03Z |

Both were also paid on eliza in batch 1 (`5TcZmisQav3A…` at 17:14:34Z, `4x1BdCZ6dpJt…` at 17:13:27Z). My best guess at the difference is that `getTokenAccountsByOwner` returns current state, and those accounts were closed after the funds moved out, so they are empty today and the query reads as "never existed". Worth a second look on your side, since you have the per-wallet JSON.

Only `EU8ZXW65…` (FreeSolDev's asi-registered wallet) received nothing, and that contributor was paid 25.036860 at `soAVqvYm…`, the wallet in the eliza preparation, at 00:46:28Z (`22TWb4KEFcF2…`).

If that read holds, the fixture's "203076752 unpaid" is describing a settlement that did not happen, and the test only passes because `observe()` synthesizes observations from entitlements. I would swap it for observed data. I have the 19 batch-2 transfers as JSON keyed by address with signatures and am happy to hand it over, or push it onto this branch as a fixture if that is easier. Approval the same day once the fixture describes the real settlement.

**2. What the function reports against real observations**, which is the case for merging it.

asi 2026-08, batch 2 (19 transfers, 1,785.267199 USDC): 15 `exact`, 2 `mismatch` (ss251 +5.563747, rama0x1 +2.781873, both exact plus a multiple of the 10,000-weight share), 1 `unpaid` (FreeSolDev's asi wallet, above), 8 `unpayable`, 2 `unexpectedAddresses`. Both unexpected addresses are eliza-registered wallets of asi contributors: FreeSolDev, and Svector-anu, who has no wallet in the asi preparation but received 233.677358 at `Ex6ePz4U…` (230.895485 entitlement plus one 2.781873 share). `reconciles: false`, correctly.

eliza 2026-08, batch 1 (53 transfers, 4,219.710147 USDC): 0 `exact`, 53 `mismatch`, 3 `unpaid`. The row you spotted (7466401 vs 7453416) turns out not to be a lone gap. All 52 non-mashingaan rows are short by the same factor, observed / entitled = 0.9982578 to rounding, so the batch was computed on a 9,982.58 USDC pool rather than the 10,000 cap. Same preparation, different basis. mashingaan is +12.347913, the #434 top-up. The three unpaid rows are OxBenji 1.659200, fishwishes 0.829600 and FLAVIEN66 0.829600, all under `minimumTransferMinor`.

**3. Non-blocking notes.**

- `unpaid` cannot tell a row correctly withheld under the dust floor from a row that was skipped, so a CLI that stops on any `unpaid` would fail on the three eliza rows above. A `withheld` status when `entitledMinor < minimumTransferMinor`, or the floor as an input, would keep that distinction.
- Per preparation, an address registered under another project is `unexpected`, which is right for the primitive. The CLI should probably cross-reference the other preparations before calling an address stray, since two of the four surprises in this settlement were exactly that.
- Tiny wording thing: the module comment says a preparation records what each contributor "is owed". The states we publish are projected, under review, approved, scheduled, paid, unclaimed, held. "Projected" or "entitled under the preparation" would keep the code in step with that vocabulary.

Leaving this as a comment rather than a block. The CLI you offered is wanted, and thank you for the re-review on #454 yesterday.

lalalune:
Reviewed and repaired exact head `54379103e6aee6d9a629dbfe7bf064c71c6f7c09` on current develop `9748eadb976e45db4509350d29e2c91a55eb3f22`. The primitive is read-only; it preserves mismatch, withheld and unpayable states without claiming payment authority. Added fail-closed shared-wallet handling and validation for all observed amounts.

[Hosted validation](https://github.com/SlopDotCash/slopdotcash/actions/runs/36053973216): all required checks pass; 137 files / 1,562 unit tests, all skill tests, 117 browser tests and 5 Pages contracts pass. Three duplicate reflow cases are intentionally skipped. [Browser diagnostics](https://github.com/SlopDotCash/slopdotcash/actions/runs/36053973216/artifacts/10831699955), [source/build evidence](https://github.com/SlopDotCash/slopdotcash/actions/runs/36053973216/artifacts/10831859510).

Local full verify: 1,561 passed, one held-cycle build subprocess exceeded its timeout; that test passes in the complete hosted run. Focused reconciliation: 11 passed. Production deployment, DNS, TLS, external provider and settlement execution: N/A for this library review, not claimed.

Review assistance: OpenAI GPT-6 / Codex desktop, self-reported.

PR #466, opened 2026-09-19
feat(wallet): prove possession of a payout destination key

drew-dot-com:
Closes nothing on its own. Related to the destination-integrity thread on #338 and #366.

## Why

A wallet claim binds a GitHub actor to an address and nothing else. `WalletClaim` has no signature field, and `prepareWalletRegistration` checks only that the string is a canonical 32-byte base58 key. The site is honest about the consequence in four places, including `src/App.tsx`:

> GitHub identity never proves wallet control.

So the largest unproven assumption on the payout path is that the address a contributor typed is one they can actually spend from. On Solana a transfer to a mistyped address, an address someone else controls, or an address pasted out of a poisoned transaction history is final. We have already seen the adjacent version of this: lookalike addresses dusted the payout wallet within seconds of the first August payment.

This turns that assumption into evidence a reviewer can check, without changing what registration requires.

## What it does

`slop-wallet-possession:v1` verifies an Ed25519 signature by the claimed destination key over a server-issued challenge bound to one claim.

It follows the already-reviewed `protocol/signer-access-attestations.md` contract:

- Domain-separated message whose first line states what the signature means and expressly denies payment authority, so a signature collected here cannot be presented as consent to move funds
- The signed bytes are a UTF-8 string, not a serialized transaction message
- Strict Noble Curves verification with `zip215: false`, not permissive consensus verification that accepts identity-point forgeries
- Bounded lifetime fixed by the module, so a caller cannot widen its own replay window
- Every bound field compared against server-held state, so a signature for one claim, actor, or address cannot be replayed onto another
- Challenge consumed atomically before the signature is trusted
- Verification only. **This repository still contains no signing code.** Contributors sign with their own wallet.

## Additive by construction

| | |
| --- | --- |
| Registration | Unchanged. Still no wallet connection or signature required. |
| Existing claims | Read `unproven`, the honest pre-existing state. |
| `unproven` | Never holds, excludes, reduces, or delays a position. |
| Payout path | Reads none of this. No reward, allocation, settlement, or verification code touches the new table. |
| Off-curve destinations | Report `not-provable-by-signature` rather than failing, so a contributor paid to a Squads vault is not pushed toward a worse destination to satisfy a check. |

The immutable by-id claim response is deliberately unchanged. A claim is immutable; an attestation about it is a separate append-only record, so adding one never invalidates a response already served as `immutable`.

## Endpoints

```text
POST /api/v1/wallet-claims/<claimId>/possession-challenge   authenticated
POST /api/v1/wallet-claims/<claimId>/possession             authenticated
GET  /api/v1/wallet-claims/<claimId>/possession             public
```

## Verification

- `bun run typecheck` clean on both projects
- `bun run lint:check` clean
- Full unit suite: **126 files, 1498 tests, all passing**
- 15 new tests: 10 on the verifier, 5 end to end through the handler, covering a valid proof, a signature by a different key, replay onto another claim/actor/address, expiry at both edges, a challenge that widens its own window, single-use consumption, another contributor's claim, and the off-curve path

## Not in this PR

No UI. The signing step needs a real wallet extension to verify end to end, and I am not going to add a control to `/wallet` that I have not driven with an actual wallet in hand. That is a follow-up PR with real-browser evidence attached. The API is complete and usable in the meantime.

No migration has been applied. `migrations/0007_wallet_possession_attestations.sql` ships here; applying it to D1 is a separate operator step.


drew-dot-com:
For the record: @lalalune approved this by DM on 21 Sep 2026 before I merged it. It touches the money path, so I did not self-merge on silence. Trial-merged onto current develop first, full unit suite green.

Issue #472, opened 2026-09-20
RFC: Base as a settlement network, one network per project


drew-dot-com:
## TL;DR

Today a sponsor can commit money on Base (a Sablier stream), but a cycle can only be paid and marked `paid` on Solana. So a Base-funded pool has no path to settlement inside the pipeline. This RFC proposes letting a project declare Base as its settlement network, with the same non-custodial rules as Solana. It asks for one decision before any money-path code is written: **one settlement network per project, not per contributor.**

Nothing here changes custody. Slop still never holds funds or keys, never signs, never broadcasts. The creator signs ordinary USDC transfers from their own wallet and Slop verifies the result.

## Why now

- Base funding verification and Sablier Lockup commitments on Base already ship. The payout leg is the missing half.
- `cycles/README.md` already says it: "A pledged or Sablier/EVM basis cannot produce a Solana plan without a separately reviewed funding transition." A Base-committed pool is stuck at that sentence.
- Base is reviewing our Batches application. A Base-funded cycle closed end to end is the most useful thing we could show them, and it is also simply the correct way to pay out a Base-funded pool.
- #470 adds the read-only reconciliation check for a Base payout transaction. It writes nothing and does not depend on this RFC.

## What is Solana-only today

- Wallet claim marker: `{"chain":"solana","address":...}` is the only accepted form (`src/lib/wallets.ts`).
- Allocation, payout wallet, and settlement types carry a literal `chain: "solana"` (`src/lib/rewards.ts`).
- Execution plan: kind `solana-usdc-transfer-plan`, SPL mint, Solana Pay requests (`src/lib/settlement-plan.ts`).
- Evidence and verification: `solana-settlement-evidence`, Solana RPC (`scripts/verify-settlement.ts`).

## Proposal

1. `projects/<id>/project.json` gains `settlementNetwork`, `"solana"` by default, `"base"` allowed. Changed by PR like everything else, and only between cycles. A frozen proposal keeps the network it was frozen with.
2. Wallet claims accept `{"chain":"base","address":"0x..."}` with lowercase canonical addresses. A contributor may hold one claim per chain. Existing rules carry over unchanged: a claim observed after proposal generation applies to the next cycle, history is never edited.
3. A row whose contributor has no wallet on the project's settlement network stays `unclaimed`, exactly as a missing wallet does today.
[...]

PR #478, opened 2026-09-21
feat(wallets): accept Base wallet claims, one lineage per chain (draft, waits on #472)

drew-dot-com:
## TL;DR

Lets a contributor register a Base wallet next to their Solana wallet. Nothing pays to it yet. This is step 2 of the sequence in #472 and it should not merge until #472 is answered. It is a draft so the design is on the table, not a request to ship.

Money path, so this is @lalalune's to merge.

## What changes

- `src/lib/wallets.ts`: wallet markers may name `chain: "base"` with a lowercase `0x` address. At most one marker per chain. A second marker on the same chain still fails closed, so an old address can never stay ambiguously payable. `parsePublishedWallet` still returns only the Solana marker, so every existing caller behaves exactly as before.
- `migrations/0007_wallet_claim_chain.sql`: adds `chain` to `wallet_claims` (default `solana`), makes the one-root index per actor per chain, and adds an insert trigger so a successor must stay on its predecessor's chain and actor. A Base claim can never retire a Solana address.
- `backend/trace`: `POST /wallet-claims` takes an optional `chain`. Both `current` reads take `?chain=`. With no `chain`, everything resolves Solana, so the reward pipeline and the registration page are untouched.
- Solana records keep their original digested byte layout. Only Base records include `chain` in the digest. There is a test that recomputes a Solana digest from the old layout.
- The operator recovery endpoint stays Solana-only.

## What this deliberately does not do

- No settlement, plan, allocation, or verifier change. `WalletProof` is still `chain: "solana"`. A Base claim is inert until `settlementNetwork` exists (step 3 in #472).
- No UI. Merge to develop is the release, and offering a Base option on the live page before a Base cycle can pay would invite claims that then sit `unclaimed` with no explanation. The UI belongs with step 3.
- No change to the three contributor skills' `wallet-claim.mjs`, so no skill revision changes.
- No EIP-55 checksum. Lowercase only gives one spelling per account, which keeps digests and duplicate checks exact. Easy to revisit if you would rather accept and normalize checksummed input.

## Verification

- `vitest run`: 126 files, 1497 tests pass. `typecheck`, `check:unused`, `protocol:check`, biome clean. I did not run the full `bun run verify` or the browser E2E; there is no UI change.
- Migration applied to a real SQLite database after 0001 to 0006 with an existing Solana row present. The existing row reads back as `solana`. A second root on the same chain, a Base successor to a Solana claim, a successor under a different actor, a fork, and an unknown chain are all rejected by the database. A Base root beside a Solana root and a Base successor are accepted. `ADD COLUMN` does not fire the no-update trigger.
- Not tested against the production D1 instance.

## Questions

1. Lowercase only, or accept EIP-55 and normalize?
2. Is `?chain=` on the existing routes fine, or would you rather have a separate route per chain?

Disclosure: I am also an equity partner at TOON Protocol. This has no connection to TOON.


lalalune:
Review of draft head `1a988431acf3bf4d1a23c440838e066186e6019c`: #472 remains unanswered. The branch also conflicts with the subsequently merged wallet-possession API. Its migration number 0007 now collides with wallet-possession migration 0007; the current sequence already reaches 0009.

I prepared a local integration repair that keeps both APIs, renames the new migration to 0010, and explicitly rejects Base claims at the Solana-only possession endpoints instead of treating Base addresses as Ed25519 destinations. The existing 60 focused wallet/API tests pass, and added checks confirm all three proof endpoints return unsupported_chain for Base. A SQLite upgrade with an existing Solana claim passes and rejects duplicate roots, cross-chain/cross-user successors, forks, unknown chains, and updates/deletes.

The repaired branch has not been pushed or marked ready while the requested RFC/scope decision is pending. No production D1 migration or settlement change was performed. Full exact-head CI remains required after that decision and integration.

Review assistance: OpenAI GPT-6 / Codex desktop, self-reported.

lalalune (approved):
Reviewed exact head `e36b4f9e4e43143a788f796f107f7d032276223a`. The maintainer explicitly authorized finishing #478, resolving the draft hold. Lowercase Base addresses and the optional chain selector are retained; Solana remains the default only when the field is omitted.

Repaired the migration collision (now 0010), preserved the merged possession endpoints, rejected Base claims at all Solana proof endpoints, and added a real SQLite upgrade/invariant regression. Sixty-one focused tests pass. The full hosted run passes 1,594 tests across 139 files, skill contracts, type/lint/build gates, 119 preview browser tests and five Pages tests, plus every trusted transition gate: https://github.com/SlopDotCash/slopdotcash/actions/runs/36067992655 . Source bundles and browser diagnostics are attached there.

The local full verification attempt encountered an integration packaging timeout under machine load and is not reported passing; hosted validation completed independently. No UI change, production D1 apply, Base settlement, fee recipient, or payment is included or claimed. Production deployment evidence is N/A for this source merge.

Review assistance: OpenAI GPT-6 / Codex desktop, self-reported.

